PR #24, opened 2026-04-18
Add extensible structured client logging to the Sunlight framework

### LINKED ISSUE #11: Add extensible structured client logging to the NScript framework
## Summary
Add framework-level structured client logging to NScript with an extensible sink/transport model so browser-side log events can be captured as structured payloads and forwarded beyond the console.

## Value
This gives NScript apps a real client logging story: developers and operators can route client events into server-side pipelines such as NLog or Serilog, preserve category/correlation metadata, and extend transports over time without rewriting call sites.

**Priority:** Medium

## Acceptance Criteria
- [ ] Add a framework-level logging abstraction for structured log events instead of console-only emission.
- [ ] Support logger/category/module identity so log records can be attributed to the calling component.
- [ ] Provide a first transport or sink implementation that sends structured log events over HTTP.
- [ ] Keep the transport model extensible so future sinks such as WebSocket can be added without redesigning the public API.
- [ ] Define a stable structured payload/schema suitable for server-side ingestion into Serilog or NLog.
- [ ] Preserve correlation metadata where available and keep existing structured context behavior usable.
- [ ] Add a reference server-side ingestion example in `TestWebApplication` showing how received client events can be written as structured logs on the server.
- [ ] Preserve a console sink/output option for local diagnostics and backward-friendly adoption.
- [ ] Document how framework consumers create named/category loggers and configure HTTP transport usage.

gautam-achieveai:
Closes #11.

Refactors the client-side `Logger` into a sink-based pipeline while keeping the pre-WI-11 API (`Logger.Debug`/`Info`/`Warn`/`Error`) fully backwards compatible — existing callers such as `TaskScheduler`'s exception handlers compile unchanged.

## Summary

- **`ILogSink`** contract with `Handle`/`Flush`/`Detach` lifecycle.
- **`ConsoleSink`** — default lazy-installed sink that writes one-line JSON to the appropriate `console.*` method.
- **`HttpLogSink`** — batched HTTP transport with size/timer flush, bounded queue with oldest-drop overflow, and `navigator.sendBeacon` fallback on `pagehide`/`beforeunload` for reliable unload delivery.
- **`ILogger` / `NamedLogger`** — category-scoped loggers via `Logger.ForCategory("TodoApp.ListView")`. Instance-cached so holding a static ref is zero-allocation per call.
- **`LogEvent`** immutable snapshot auto-enriched with `CallContext` correlation (`traceId`, `spanId`, `parentSpanId`, `actionId`, `depth`).
- **`LogJsonBuilder`** — pure-C# JSON envelope assembly that survives NScript minification; only string-escape defers to the browser's `JSON.stringify`.

## Compile-time stripping

`Logger.Trace`/`Debug` and `NamedLogger.Trace`/`Debug` carry `[Conditional("DEBUG")]`, so Release builds strip call sites (including argument evaluation) via Roslyn. `Trace`/`Debug` are intentionally **not** on `ILogger` because C# disallows `[Conditional]` on interface members — exposing them there would silently bypass stripping.

## Reliability

- Per-sink `try/catch` in `Logger.DispatchInternal`, `Logger.Flush`, `Logger.RemoveSink`, `Logger.ClearSinks` so a faulty sink cannot break its peers or the caller (which may be an exception handler).
- Re-entrancy guards in `Logger.DispatchInternal` (`dispatching` flag) and `HttpLogSink.Flush` (`flushing` flag) prevent duplicate delivery and runaway recursion.
- `HttpLogSink` swallows transport errors in `Flush`/`Detach`/`OnPageUnload` — the timer callback and unload handler cannot meaningfully react to a network failure, and the design is documented as fire-and-forget.

## JSON schema

```json
{
  "ts":       "2026-04-17T18:00:00.000Z",
  "level":    "INFO",
  "cat":      "TodoApp.ListView",
  "msg":      "Item added",
  "props":    { "itemId": "42" },
  "traceId":       "a1b2c3d4e5f6a7b8a1b2c3d4e5f6a7b8",
  "spanId":        "a1b2c3d4e5f6a7b8",
  "parentSpanId":  "f1e2d3c4b5a6f7e8",
  "actionId":      5,
  "depth":         1
}
```

HTTP envelope: `{ "events": [...], "dropped": N }`.

## Test plan

- [x] 17 QUnit framework tests in `LoggerTests.cs` — level filtering, multi-sink fan-out, category carry, instance caching, null-category normalization, `CallContext` enrichment, property passthrough, sink fault isolation, `RemoveSink`/`Detach`, `ClearSinks` suppresses lazy default, `Flush`, `Trace` dispatches under DEBUG, `NamedLogger.Trace` carries category, `HttpLogSink` batch-on-size / timer-cleared-on-detach / overflow-drops-oldest.
- [x] All 168 framework QUnit tests pass (759/759 assertions).
- [x] Debug + Release builds clean (0 errors).
- [x] Existing callers (`TaskScheduler`) compile unchanged — confirmed no `Logger.Emit` references anywhere.

## Documentation

`docs/framework-logging.md` covers the JSON schema, server-side wiring snippets for Serilog and NLog, compile-time stripping semantics, lifecycle patterns, and out-of-scope items for follow-up (WebSocket transport, per-category level overrides, sampling, message templates, localStorage retry).

## Reference ingestion handler

`Test/Framework/TestWebApplication/ClientLogs.ashx(.cs)` — minimal handler that forwards each batched event to `Trace.WriteLine`. Intended as a starting point; docs show how to plug into Serilog / NLog / AppInsights.

### DIFF
--- a/Sources/Framework/Sunlight.Framework/Logger.cs
+++ b/Sources/Framework/Sunlight.Framework/Logger.cs
@@ -216,10 +216,11 @@ namespace Sunlight.Framework
         /// caller, which may be an exception handler).
         /// </summary>
         /// <remarks>
-        /// Internal (not public) because the facade and <see cref="NamedLogger"/>
-        /// already apply the <see cref="MinLevel"/> check. External callers
-        /// should always go through the level-specific methods so level filters
-        /// and <c>[Conditional]</c> stripping are honored.
+        /// Internal (not public) because the static facade already applies the
+        /// <see cref="MinLevel"/> check; <see cref="NamedLogger"/> must apply it
+        /// too to match that contract. External callers should always go through
+        /// the level-specific methods so level filters and <c>[Conditional]</c>
+        /// stripping are honored.
         /// </remarks>
         internal static void DispatchInternal(
             LogLevel level,
@@ -261,6 +262,15 @@ namespace Sunlight.Framework
                     try { Logger.sinks[i].Handle(evt); }
                     catch { /* sink faults must not escape — protects exception-path callers */ }
                 }
+            }
+            catch
+            {
+                // Event construction failed (e.g. GetIsoTimestamp bridge threw,
+                // ConsoleSink allocation failed). Swallowing is the only safe
+                // action: Logger.Error() is called from inside TaskScheduler's
+                // catch blocks, so an escaping exception would bypass its
+                // finally cleanup (currentTask / CallContext restore) and
+                // produce cascading failures.
             }
             finally
             {
--- a/Sources/Framework/Sunlight.Framework/Logger.cs
+++ b/Sources/Framework/Sunlight.Framework/Logger.cs
@@ -7,24 +7,39 @@
 namespace Sunlight.Framework
 {
     using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
     using System.Runtime.CompilerServices;
 
-    public enum LogLevel
-    {
-        Debug = 0,
-        Info = 1,
-        Warn = 2,
-        Error = 3
-    }
-
     /// <summary>
-    /// Structured client-side logger. Every entry is auto-enriched with
-    /// CallContext IDs (actionId, traceId, spanId) for OTEL correlation.
-    /// Output goes to browser console as structured JSON.
+    /// Static entry point for structured client-side logging. Fans emitted
+    /// <see cref="LogEvent"/>s out to every registered <see cref="ILogSink"/>
+    /// after <see cref="MinLevel"/> filtering.
     /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Backwards-compatible with the pre-WI-11 API: the static
+    /// <see cref="Debug"/> / <see cref="Info"/> / <see cref="Warn"/> /
+    /// <see cref="Error"/> methods still accept a single <c>string</c> message.
+    /// Pre-existing callers (e.g. <c>TaskScheduler</c>'s exception handlers) do
+    /// not need to change.
+    /// </para>
+    /// <para>
+    /// <see cref="Trace"/> and <see cref="Debug"/> are tagged with
+    /// <c>[Conditional("DEBUG")]</c>: their calls — including argument
+    /// evaluation — are stripped from the bound tree by Roslyn during Stage 1
+    /// of the NScript pipeline whenever the caller's compilation is built
+    /// without the <c>DEBUG</c> symbol (i.e. the Release config). This yields
+    /// zero runtime overhead for verbose logs in production builds.
+    /// </para>
+    /// </remarks>
     public static class Logger
     {
         private static LogLevel minLevel = LogLevel.Debug;
+        private static List<ILogSink> sinks;
+        private static StringDictionary<NamedLogger> categoryCache;
+        private static bool userConfigured;
+        private static bool dispatching;
 
         public static LogLevel MinLevel
         {
@@ -32,60 +47,233 @@ namespace Sunlight.Framework
             set { Logger.minLevel = value; }
         }
 
+        /// <summary>
+        /// Obtain (and cache) a category-scoped logger. Subsequent calls with
+        /// the same category string return the same instance so consumers can
+        /// hold a cached reference at static-init time without worrying about
+        /// allocations per call site.
+        /// </summary>
+        /// <remarks>
+        /// Return type is the concrete <see cref="NamedLogger"/>, not
+        /// <see cref="ILogger"/>: the <c>[Conditional("DEBUG")]</c> on Trace /
+        /// Debug only fires through the concrete type. See <see cref="ILogger"/>
+        /// for the full rationale.
+        /// </remarks>
+        public static NamedLogger ForCategory(string category)
+        {
+            if (category == null) { category = string.Empty; }
+            if (Logger.categoryCache == null)
+            {
+                Logger.categoryCache = new StringDictionary<NamedLogger>();
+            }
+
+            NamedLogger existing;
+            if (Logger.categoryCache.TryGetValue(category, out existing))
+            {
+                return existing;
+            }
+
+            var created = new NamedLogger(category);
+            Logger.categoryCache[category] = created;
+            return created;
+        }
+
+        public static void AddSink(ILogSink sink)
+        {
+            if (sink == null) { return; }
+            Logger.EnsureSinkList();
+            Logger.userConfigured = true;
+            Logger.sinks.Add(sink);
+        }
+
+        public static void RemoveSink(ILogSink sink)
+        {
+            if (sink == null || Logger.sinks == null) { return; }
+            if (Logger.sinks.Remove(sink))
+            {
+                try { sink.Detach(); }
+                catch { /* sink cleanup must not surface to caller */ }
+            }
+        }
+
+        /// <summary>
+        /// Remove every registered sink and detach its resources. After this
+        /// call no sinks are installed — the default lazy <see cref="ConsoleSink"/>
+        /// is suppressed until the caller explicitly adds one via
+        /// <see cref="AddSink"/>. This makes the silent-mode setup possible.
+        /// </summary>
+        public static void ClearSinks()
+        {
+            Logger.userConfigured = true;
+            if (Logger.sinks == null) { return; }
+            for (int i = 0; i < Logger.sinks.Count; i++)
+            {
+                try { Logger.sinks[i].Detach(); }
+                catch { /* sink cleanup must not surface to caller */ }
+            }
+            Logger.sinks.Clear();
+        }
+
+        /// <summary> Flush every registered sink. No-op for sinks without buffering. </summary>
+        public static void Flush()
+        {
+            if (Logger.sinks == null) { return; }
+            for (int i = 0; i < Logger.sinks.Count; i++)
+            {
+                try { Logger.sinks[i].Flush(); }
+                catch { /* flush failure must not surface to caller */ }
+            }
+        }
+
+        [Conditional("DEBUG")]
+        public static void Trace(string message)
+        {
+            if (Logger.minLevel <= LogLevel.Trace)
+            {
+                Logger.DispatchInternal(LogLevel.Trace, string.Empty, message, null);
+            }
+        }
+
+        [Conditional("DEBUG")]
+        public static void Trace(string message, string[] properties)
+        {
+            if (Logger.minLevel <= LogLevel.Trace)
+            {
+                Logger.DispatchInternal(LogLevel.Trace, string.Empty, message, properties);
+            }
+        }
+
+        [Conditional("DEBUG")]
         public static void Debug(string message)
         {
             if (Logger.minLevel <= LogLevel.Debug)
-                Logger.Emit(LogLevel.Debug, message);
+            {
+                Logger.DispatchInternal(LogLevel.Debug, string.Empty, message, null);
+            }
+        }
+
+        [Conditional("DEBUG")]
+        public static void Debug(string message, string[] properties)
+        {
+            if (Logger.minLevel <= LogLevel.Debug)
+            {
+                Logger.DispatchInternal(LogLevel.Debug, string.Empty, message, properties);
+            }
         }
 
         public static void Info(string message)
         {
             if (Logger.minLevel <= LogLevel.Info)
-                Logger.Emit(LogLevel.Info, message);
+            {
+                Logger.DispatchInternal(LogLevel.Info, string.Empty, message, null);
+            }
+        }
+
+        public static void Info(string message, string[] properties)
+        {
+            if (Logger.minLevel <= LogLevel.Info)
+            {
+                Logger.DispatchInternal(LogLevel.Info, string.Empty, message, properties);
+            }
         }
 
         public static void Warn(string message)
         {
             if (Logger.minLevel <= LogLevel.Warn)
-                Logger.Emit(LogLevel.Warn, message);
+            {
+                Logger.DispatchInternal(LogLevel.Warn, string.Empty, message, null);
+            }
+        }
+
+        public static void Warn(string message, string[] properties)
+        {
+            if (Logger.minLevel <= LogLevel.Warn)
+            {
+                Logger.DispatchInternal(LogLevel.Warn, string.Empty, message, properties);
+            }
         }
 
         public static void Error(string message)
         {
             if (Logger.minLevel <= LogLevel.Error)
-                Logger.Emit(LogLevel.Error, message);
+            {
+                Logger.DispatchInternal(LogLevel.Error, string.Empty, message, null);
+            }
+        }
+
+        public static void Error(string message, string[] properties)
+        {
+            if (Logger.minLevel <= LogLevel.Error)
+            {
+                Logger.DispatchInternal(LogLevel.Error, string.Empty, message, properties);
+            }
         }
 
         /// <summary>
-        /// Emits a structured JSON log entry to the browser console, enriched
-        /// with CallContext IDs when an action context is active.
+        /// Shared dispatch path used by the static facade and <see cref="NamedLogger"/>.
+        /// Builds a <see cref="LogEvent"/> and fans out to every sink under a
+        /// try/catch so a single faulty sink cannot break the others (or the
+        /// caller, which may be an exception handler).
         /// </summary>
         /// <remarks>
-        /// [Script] is required here because this method uses JS-native APIs that
-        /// have no NScript C# facade: Date.now(), JSON.stringify(), console.log/warn/error,
-        /// and dynamic property assignment on an object literal (entry.actionId = ...).
-        /// None of these are expressible in NScript's C# subset.
+        /// Internal (not public) because the facade and <see cref="NamedLogger"/>
+        /// already apply the <see cref="MinLevel"/> check. External callers
+        /// should always go through the level-specific methods so level filters
+        /// and <c>[Conditional]</c> stripping are honored.
         /// </remarks>
-        [Script(@"
-            var entry = {
-                ts: Date.now(),
-                level: level === 0 ? 'DEBUG' : level === 1 ? 'INFO' : level === 2 ? 'WARN' : 'ERROR',
-                msg: message
-            };
-            var ctx = @{[Sunlight.Framework]Sunlight.Framework.CallContext::current};
-            if (ctx) {
-                entry.actionId = ctx.@{[Sunlight.Framework]Sunlight.Framework.CallContext::ActionId};
-                entry.traceId = ctx.@{[Sunlight.Framework]Sunlight.Framework.CallContext::TraceId};
-                entry.spanId = ctx.@{[Sunlight.Framework]Sunlight.Framework.CallContext::SpanId};
-                if (ctx.@{[Sunlight.Framework]Sunlight.Framework.CallContext::ParentSpanId}) {
-                    entry.parentSpanId = ctx.@{[Sunlight.Framework]Sunlight.Framework.CallContext::ParentSpanId};
+        internal static void DispatchInternal(
+            LogLevel level,
+            string category,
+            string message,
+            string[] properties)
+        {
+            if (level < Logger.minLevel) { return; }
+
+            // Re-entrancy guard protects against a misbehaving sink that calls
+            // back into Logger.* (which would otherwise re-enter and duplicate
+            // events). Dropping the nested call is preferable to deadlocking
+            // or unbounded recursion.
+            if (Logger.dispatching) { return; }
+
+            Logger.dispatching = true;
+            try
+            {
+                Logger.EnsureSinkList();
+                if (!Logger.userConfigured && Logger.sinks.Count == 0)
+                {
+                    // Lazy default: first log call installs a ConsoleSink so
+                    // existing callers get the old on-console behavior without
+                    // any wiring. Once the app opts in (AddSink/ClearSinks)
+                    // this branch never re-fires.
+                    Logger.sinks.Add(new ConsoleSink());
+                }
+
+                var evt = new LogEvent(
+                    Logger.GetIsoTimestamp(),
+                    level,
+                    category,
+                    message,
+                    properties,
+                    CallContext.Current);
+
+                for (int i = 0; i < Logger.sinks.Count; i++)
+                {
+                    try { Logger.sinks[i].Handle(evt); }
+                    catch { /* sink faults must not escape — protects exception-path callers */ }
                 }
             }
-            var str = @:JSON.stringify(entry);
-            if (level >= 3) console.error(str);
-            else if (level >= 2) console.warn(str);
-            else console.log(str);
-        ")]
-        private static extern void Emit(LogLevel level, string message);
+            finally
+            {
+                Logger.dispatching = false;
+            }
+        }
+
+        private static void EnsureSinkList()
+        {
+            if (Logger.sinks == null) { Logger.sinks = new List<ILogSink>(); }
+        }
+
+        [Script("return new Date().toISOString();")]
+        private static extern string GetIsoTimestamp();
     }
 }
--- a/Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
+++ b/Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
@@ -38,6 +38,8 @@ namespace Sunlight.Framework
         private readonly int maxQueueSize;
         private readonly IWindowTimer timer;
 
+        private readonly Action<string, string> transportOverride;
+
         private List<LogEvent> queue;
         private int droppedCount;
         private int timerHandle;
@@ -51,6 +53,24 @@ namespace Sunlight.Framework
             int flushIntervalMs,
             int maxQueueSize,
             IWindowTimer timer)
+            : this(endpoint, batchSize, flushIntervalMs, maxQueueSize, timer, null)
+        {
+        }
+
+        /// <summary>
+        /// Test-only constructor: <paramref name="transportOverride"/> replaces
+        /// both the XHR POST and <c>sendBeacon</c> paths, so unit tests can
+        /// capture the serialized envelope without touching the network. When
+        /// <paramref name="transportOverride"/> is <c>null</c> the production
+        /// transports are used.
+        /// </summary>
+        internal HttpLogSink(
+            string endpoint,
+            int batchSize,
+            int flushIntervalMs,
+            int maxQueueSize,
+            IWindowTimer timer,
+            Action<string, string> transportOverride)
         {
             if (endpoint == null) { throw new ArgumentNullException("endpoint"); }
             if (timer == null) { throw new ArgumentNullException("timer"); }
@@ -60,6 +80,7 @@ namespace Sunlight.Framework
             this.flushIntervalMs = flushIntervalMs;
             this.maxQueueSize = maxQueueSize;
             this.timer = timer;
+            this.transportOverride = transportOverride;
 
             this.queue = new List<LogEvent>();
             this.droppedCount = 0;
@@ -77,13 +98,24 @@ namespace Sunlight.Framework
         {
             if (this.detached) { return; }
 
-            // Overflow: drop oldest events, count the drops. We bias toward
-            // keeping the *most recent* entries because those are usually what
-            // the developer wants to see when triaging a crash.
-            while (this.queue.Count >= this.maxQueueSize)
+            // Overflow: drop the oldest events in one pass, count the drops.
+            // We bias toward keeping the *most recent* entries because those
+            // are usually what the developer wants to see when triaging a
+            // crash. A previous implementation used List<T>.RemoveAt(0) in a
+            // loop, which is O(n) per drop (Array.splice(0,1) shifts every
+            // remaining element); with maxQueueSize=500 under sustained
+            // overflow that compounds badly. Rebuilding the tail into a new
+            // list is a single O(n) pass instead.
+            if (this.queue.Count >= this.maxQueueSize)
             {
-                this.queue.RemoveAt(0);
-                this.droppedCount++;
+                int excess = this.queue.Count - this.maxQueueSize + 1;
+                this.droppedCount += excess;
+                var trimmed = new List<LogEvent>();
+                for (int i = excess; i < this.queue.Count; i++)
+                {
+                    trimmed.Add(this.queue[i]);
+                }
+                this.queue = trimmed;
             }
 
             this.queue.Add(evt);
@@ -105,14 +137,16 @@ namespace Sunlight.Framework
             this.flushing = true;
             try
             {
+                // Single protection boundary covering BOTH payload construction
+                // and transport. LogJsonBuilder.BuildEnvelope (called via
+                // ExtractBatchAsPayload) can throw on JSON serialization
+                // failure; the outer timer/unload callers cannot meaningfully
+                // react to any failure here. Lost batches are the documented
+                // fire-and-forget contract.
                 string payload = this.ExtractBatchAsPayload();
-                // Transport failures must not escape — callers include the
-                // timer tick and page-unload handler, neither of which can
-                // meaningfully react to a network error. Lost batches are the
-                // documented fire-and-forget contract.
-                try { HttpLogSink.PostPayload(this.endpoint, payload); }
-                catch { /* fire-and-forget transport */ }
+                this.SendPayload(payload, false);
             }
+            catch { /* fire-and-forget: payload build or transport failure */ }
             finally
             {
                 this.flushing = false;
@@ -134,21 +168,33 @@ namespace Sunlight.Framework
             this.unloadHandler = null;
 
             // Best-effort final flush — use sendBeacon because the normal XHR
-            // path cannot be trusted to complete once we stop listening.
+            // path cannot be trusted to complete once we stop listening. The
+            // try/catch covers ExtractBatchAsPayload too, because serialization
+            // failure after the destructive queue swap would otherwise escape
+            // through RemoveSink/ClearSinks up into arbitrary caller code.
             if (this.queue.Count > 0)
             {
-                string payload = this.ExtractBatchAsPayload();
-                try { HttpLogSink.SendBeacon(this.endpoint, payload); }
+                try
+                {
+                    string payload = this.ExtractBatchAsPayload();
+                    this.SendPayload(payload, true);
+                }
                 catch { /* fire-and-forget transport */ }
             }
         }
 
         private void OnTimerTick()
         {
+            // Runs directly on window.setInterval — OUTSIDE any application
+            // try/catch. An unhandled exception here would fire every
+            // flushIntervalMs for the lifetime of the page as an unhandled
+            // setInterval error, so the outer try/catch is mandatory even
+            // though Flush() is also internally guarded.
             if (this.detached) { return; }
             if (this.queue.Count > 0)
             {
-                this.Flush();
+                try { this.Flush(); }
+                catch { /* timer callback must not throw */ }
             }
         }
 
@@ -160,12 +206,36 @@ namespace Sunlight.Framework
 
         private void OnPageUnload()
         {
+            // beforeunload/pagehide dispatchers have no application-level
+            // protection, same rationale as OnTimerTick.
             if (this.detached) { return; }
             if (this.queue.Count == 0) { return; }
 
-            string payload = this.ExtractBatchAsPayload();
-            try { HttpLogSink.SendBeacon(this.endpoint, payload); }
+            try
+            {
+                string payload = this.ExtractBatchAsPayload();
+                this.SendPayload(payload, true);
+            }
             catch { /* fire-and-forget transport — unload path cannot react */ }
+        }
+
+        /// <summary>
+        /// Dispatch the serialized payload to the appropriate transport: the
+        /// XHR path for live flushes and the <c>sendBeacon</c> path for unload
+        /// / detach. Routed through <see cref="transportOverride"/> when the
+        /// test-only constructor injected one so unit tests can capture the
+        /// envelope without touching the network.
+        /// </summary>
+        private void SendPayload(string payload, bool isUnload)
+        {
+            if (this.transportOverride != null)
+            {
+                this.transportOverride(this.endpoint, payload);
+                return;
+            }
+
+            if (isUnload) { HttpLogSink.SendBeacon(this.endpoint, payload); }
+            else { HttpLogSink.PostPayload(this.endpoint, payload); }
         }
 
         /// <summary>
--- a/Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
+++ b/Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
@@ -74,6 +74,16 @@ namespace Sunlight.Framework
         {
             if (endpoint == null) { throw new ArgumentNullException("endpoint"); }
             if (timer == null) { throw new ArgumentNullException("timer"); }
+            // Numeric guards: zero/negative values would make the sink silently
+            // pathological — a zero batchSize flushes on every event, zero
+            // flushIntervalMs makes setInterval a tight loop, and zero
+            // maxQueueSize means Handle() always drops. Fail fast at ctor time
+            // instead of producing confusing runtime behavior. NScript's
+            // mscorlib does not expose ArgumentOutOfRangeException, so we use
+            // ArgumentException with a descriptive parameter name.
+            if (batchSize < 1) { throw new ArgumentException("batchSize must be >= 1"); }
+            if (flushIntervalMs < 1) { throw new ArgumentException("flushIntervalMs must be >= 1"); }
+            if (maxQueueSize < 1) { throw new ArgumentException("maxQueueSize must be >= 1"); }
 
             this.endpoint = endpoint;
             this.batchSize = batchSize;
@@ -239,18 +249,25 @@ namespace Sunlight.Framework
         }
 
         /// <summary>
-        /// Atomically swap the queue + <c>droppedCount</c> out, reset both, and
-        /// return the JSON envelope. Shared by <see cref="Flush"/>,
+        /// Serialize the current queue + <c>droppedCount</c> into a JSON
+        /// envelope, then reset both. Shared by <see cref="Flush"/>,
         /// <see cref="Detach"/>, and <see cref="OnPageUnload"/> so the three
         /// paths stay in sync and cannot drift.
         /// </summary>
+        /// <remarks>
+        /// Serialization happens BEFORE the queue is reset so that a
+        /// <c>BuildEnvelope</c> throw leaves the buffered events in place for
+        /// the next flush attempt. Resetting first would silently drop an
+        /// entire batch on any JSON build failure.
+        /// </remarks>
         private string ExtractBatchAsPayload()
         {
             var batch = this.queue;
             int dropped = this.droppedCount;
+            string payload = LogJsonBuilder.BuildEnvelope(batch, dropped);
             this.queue = new List<LogEvent>();
             this.droppedCount = 0;
-            return LogJsonBuilder.BuildEnvelope(batch, dropped);
+            return payload;
         }
 
         /// <summary>
--- a/Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
+++ b/Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
@@ -1,0 +1,282 @@
+//-----------------------------------------------------------------------
+// <copyright file="HttpLogSink.cs" company="">
+//     Copyright (c) . All rights reserved.
+// </copyright>
+//-----------------------------------------------------------------------
+
+namespace Sunlight.Framework
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Runtime.CompilerServices;
+    using System.Web;
+    using System.Web.Html;
+
+    /// <summary>
+    /// Batching HTTP transport sink. Queues <see cref="LogEvent"/>s in memory
+    /// and POSTs them to a configured endpoint when any of these triggers fire:
+    /// <list type="bullet">
+    ///   <item>queue length reaches <c>batchSize</c></item>
+    ///   <item>the injected <see cref="IWindowTimer"/> fires every <c>flushIntervalMs</c></item>
+    ///   <item><see cref="Flush"/> is called explicitly</item>
+    ///   <item>the page transitions to <c>beforeunload</c> / <c>pagehide</c>
+    ///         (uses <c>navigator.sendBeacon</c> for reliable unload delivery)</item>
+    /// </list>
+    /// </summary>
+    /// <remarks>
+    /// Overflow strategy: when the queue exceeds <c>maxQueueSize</c> the oldest
+    /// events are dropped, and a running dropped-count is stamped into the next
+    /// batch envelope (payload, not an HTTP header — sendBeacon cannot set
+    /// custom headers and custom headers would trigger CORS preflight on the
+    /// normal XHR path).
+    /// </remarks>
+    public class HttpLogSink : ILogSink
+    {
+        private readonly string endpoint;
+        private readonly int batchSize;
+        private readonly int flushIntervalMs;
+        private readonly int maxQueueSize;
+        private readonly IWindowTimer timer;
+
+        private List<LogEvent> queue;
+        private int droppedCount;
+        private int timerHandle;
+        private bool flushing;
+        private bool detached;
+        private Action unloadHandler;
+
+        public HttpLogSink(
+            string endpoint,
+            int batchSize,
+            int flushIntervalMs,
+            int maxQueueSize,
+            IWindowTimer timer)
+        {
+            if (endpoint == null) { throw new ArgumentNullException("endpoint"); }
+            if (timer == null) { throw new ArgumentNullException("timer"); }
+
+            this.endpoint = endpoint;
+            this.batchSize = batchSize;
+            this.flushIntervalMs = flushIntervalMs;
+            this.maxQueueSize = maxQueueSize;
+            this.timer = timer;
+
+            this.queue = new List<LogEvent>();
+            this.droppedCount = 0;
+            this.timerHandle = -1;
+            this.flushing = false;
+            this.detached = false;
+
+            // Schedule the periodic flush and install a best-effort unload hook
+            // so buffered events survive page navigation.
+            this.timerHandle = timer.SetInterval(this.OnTimerTick, flushIntervalMs);
+            this.InstallUnloadHandler();
+        }
+
+        public void Handle(LogEvent evt)
+        {
+            if (this.detached) { return; }
+
+            // Overflow: drop oldest events, count the drops. We bias toward
+            // keeping the *most recent* entries because those are usually what
+            // the developer wants to see when triaging a crash.
+            while (this.queue.Count >= this.maxQueueSize)
+            {
+                this.queue.RemoveAt(0);
+                this.droppedCount++;
+            }
+
+            this.queue.Add(evt);
+
+            if (this.queue.Count >= this.batchSize)
+            {
+                this.Flush();
+            }
+        }
+
+        public void Flush()
+        {
+            // Re-entrancy guard: if a downstream sink or the XHR OnBeforeSend
+            // hook triggers another log call that flushes, we would otherwise
+            // double-send the same events.
+            if (this.flushing) { return; }
+            if (this.queue.Count == 0) { return; }
+
+            this.flushing = true;
+            try
+            {
+                string payload = this.ExtractBatchAsPayload();
+                // Transport failures must not escape — callers include the
+                // timer tick and page-unload handler, neither of which can
+                // meaningfully react to a network error. Lost batches are the
+                // documented fire-and-forget contract.
+                try { HttpLogSink.PostPayload(this.endpoint, payload); }
+                catch { /* fire-and-forget transport */ }
+            }
+            finally
+            {
+                this.flushing = false;
+            }
+        }
+
+        public void Detach()
+        {
+            if (this.detached) { return; }
+            this.detached = true;
+
+            if (this.timerHandle >= 0)
+            {
+                this.timer.ClearInterval(this.timerHandle);
+                this.timerHandle = -1;
+            }
+
+            this.RemoveUnloadHandler(this.unloadHandler);
+            this.unloadHandler = null;
+
+            // Best-effort final flush — use sendBeacon because the normal XHR
+            // path cannot be trusted to complete once we stop listening.
+            if (this.queue.Count > 0)
+            {
+                string payload = this.ExtractBatchAsPayload();
+                try { HttpLogSink.SendBeacon(this.endpoint, payload); }
+                catch { /* fire-and-forget transport */ }
+            }
+        }
+
+        private void OnTimerTick()
+        {
+            if (this.detached) { return; }
+            if (this.queue.Count > 0)
+            {
+                this.Flush();
+            }
+        }
+
+        private void InstallUnloadHandler()
+        {
+            this.unloadHandler = this.OnPageUnload;
+            this.AddUnloadHandler(this.unloadHandler);
+        }
+
+        private void OnPageUnload()
+        {
+            if (this.detached) { return; }
+            if (this.queue.Count == 0) { return; }
+
+            string payload = this.ExtractBatchAsPayload();
+            try { HttpLogSink.SendBeacon(this.endpoint, payload); }
+            catch { /* fire-and-forget transport — unload path cannot react */ }
+        }
+
+        /// <summary>
+        /// Atomically swap the queue + <c>droppedCount</c> out, reset both, and
+        /// return the JSON envelope. Shared by <see cref="Flush"/>,
+        /// <see cref="Detach"/>, and <see cref="OnPageUnload"/> so the three
+        /// paths stay in sync and cannot drift.
+        /// </summary>
+        private string ExtractBatchAsPayload()
+        {
+            var batch = this.queue;
+            int dropped = this.droppedCount;
+            this.queue = new List<LogEvent>();
+            this.droppedCount = 0;
+            return LogJsonBuilder.BuildEnvelope(batch, dropped);
+        }
+
+        /// <summary>
+        /// Async XHR POST via the <see cref="XMLHttpRequest"/> facade. Implemented
+        /// in plain C# rather than a <c>[Script]</c> block because the NScript JS
+        /// parser rejects <c>new XMLHttpRequest()</c> inside inline scripts.
+        /// </summary>
+        /// <remarks>
+        /// Deliberately uses the raw instance methods <c>Open</c> / <c>Send</c>,
+        /// not the static <c>GetRaw</c> / <c>PostRaw</c> helpers — those helpers
+        /// invoke <c>XMLHttpRequest.OnBeforeSend</c>, which in turn calls back
+        /// into <c>Logger.*</c> for traceparent injection. Routing log transport
+        /// through that hook would be a latent recursion risk. Fire-and-forget;
+        /// network errors surface asynchronously on <c>onerror</c>, which we do
+        /// not hook (logging errors about logging would recurse).
+        /// </remarks>
+        private static void PostPayload(string endpoint, string payload)
+        {
+            var xhr = new XMLHttpRequest();
+            xhr.Open("POST", endpoint, true);
+            xhr.SetRequestHeader("Content-Type", "application/json");
+            xhr.Send(payload);
+        }
+
+        /// <summary>
+        /// Best-effort send during page unload. Prefers <c>navigator.sendBeacon</c>
+        /// with a <c>Blob</c> of type <c>application/json</c> — sending a bare
+        /// string defaults to <c>text/plain</c>, which ingestion servers tend to
+        /// reject. Falls back to a synchronous XHR on browsers without
+        /// <c>sendBeacon</c> so buffered events still have a chance to ship.
+        /// </summary>
+        private static void SendBeacon(string endpoint, string payload)
+        {
+            var navigator = Window.Instance.Navigator;
+            if (navigator != null && navigator.HasSendBeaconApi)
+            {
+                var opts = new BlobCreateOptions();
+                opts.Type = "application/json";
+                var blob = new Blob(payload, opts);
+                HttpLogSink.NavigatorSendBeaconBlob(endpoint, blob);
+                return;
+            }
+
+            // Very old browsers: sync XHR is the last-resort unload path.
+            var xhr = new XMLHttpRequest();
+            xhr.Open("POST", endpoint, false);
+            xhr.SetRequestHeader("Content-Type", "application/json");
+            xhr.Send(payload);
+        }
+
+        /// <summary>
+        /// Single-expression <c>[Script]</c> bridge so we can pass a <c>Blob</c>
+        /// to <c>navigator.sendBeacon</c> — the <see cref="Navigator"/> facade
+        /// only exposes <c>string</c>/<c>FormData</c> overloads.
+        /// </summary>
+        [Script(@"window.navigator.sendBeacon(endpoint, blob);")]
+        private static extern void NavigatorSendBeaconBlob(string endpoint, Blob blob);
+
+        private void AddUnloadHandler(Action handler)
+        {
+            HttpLogSink.WindowAddUnloadHandler(handler);
+        }
+
+        private void RemoveUnloadHandler(Action handler)
+        {
+            if (handler == null) { return; }
+            HttpLogSink.WindowRemoveUnloadHandler(handler);
+        }
+
+        /// <summary>
+        /// Register <paramref name="handler"/> for both <c>pagehide</c> and
+        /// <c>beforeunload</c>. Separate single-statement <c>[Script]</c> shims
+        /// keep each body small and parseable by the NScript JS frontend.
+        /// </summary>
+        [Script(@"window.addEventListener('pagehide', handler, false);")]
+        private static extern void WindowAddPageHide(Action handler);
+
+        [Script(@"window.addEventListener('beforeunload', handler, false);")]
+        private static extern void WindowAddBeforeUnload(Action handler);
+
+        [Script(@"window.removeEventListener('pagehide', handler, false);")]
+        private static extern void WindowRemovePageHide(Action handler);
+
+        [Script(@"window.removeEventListener('beforeunload', handler, false);")]
+        private static extern void WindowRemoveBeforeUnload(Action handler);
+
+        private static void WindowAddUnloadHandler(Action handler)
+        {
+            HttpLogSink.WindowAddPageHide(handler);
+            HttpLogSink.WindowAddBeforeUnload(handler);
+        }
+
+        private static void WindowRemoveUnloadHandler(Action handler)
+        {
+            HttpLogSink.WindowRemovePageHide(handler);
+            HttpLogSink.WindowRemoveBeforeUnload(handler);
+        }
+    }
+}
--- a/Sources/Framework/Sunlight.Framework/Logging/LogJsonBuilder.cs
+++ b/Sources/Framework/Sunlight.Framework/Logging/LogJsonBuilder.cs
@@ -76,6 +76,9 @@ namespace Sunlight.Framework
 
             // Flat [k,v,k,v,...] → {k:v,k:v,...}. Odd trailing element is
             // dropped: a key without a value is almost always a caller bug.
+            // Null keys are skipped (not emitted as "null":...) — a null key
+            // would produce invalid-looking JSON-object semantics and is
+            // almost certainly an accidental slot in the caller's string[].
             var props = evt.Properties;
             if (props != null && props.Length >= 2)
             {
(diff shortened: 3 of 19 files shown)


Sources/Framework/Sunlight.Framework/Logger.cs
```diff
@@ -7,85 +7,273 @@
...
+            if (level < Logger.minLevel) { return; }
+
+            // Re-entrancy guard protects against a misbehaving sink that calls
+            // back into Logger.* (which would otherwise re-enter and duplicate
+            // events). Dropping the nested call is preferable to deadlocking
+            // or unbounded recursion.
+            if (Logger.dispatching) { return; }
+
+            Logger.dispatching = true;
+            try
```
gautam-achieveai, line 240: **[Gautam's review bot] [HIGH] Missing outer catch — exception can escape `Logger.Error()` and crash callers**

The outer `try` block has a `finally` (correctly resetting `dispatching`) but **no `catch`**. Six statements execute between `dispatching = true` and the per-sink try/catch loop:

1. `EnsureSinkList()` — allocation can throw
2. `new ConsoleSink()` — allocation can throw
3. `GetIsoTimestamp()` — extern JS bridge (`new Date().toISOString()`) can throw
4. `new LogEvent(...)` — constructor allocation can throw

If any of these throw, the exception propagates directly to the caller. **`TaskScheduler` calls `Logger.Error()` from inside `catch` blocks** (TaskScheduler.cs:263, :417). An exception escaping here would bypass TaskScheduler's finally blocks that restore `CallContext.Current` and clear `currentTask`.

**Fix**: Add `catch` to the outer try:
```csharp
try
{
    // ... existing code ...
}
catch
{
    // Event construction failed (e.g. GetIsoTimestamp bridge threw).
    // Swallowing is the only safe action — Logger is called from
    // exception handlers (TaskScheduler) where throwing would be fatal.
}
finally
{
    Logger.dispatching = false;
}
```

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
```diff
@@ -0,0 +1,282 @@
...
+            // path cannot be trusted to complete once we stop listening.
+            if (this.queue.Count > 0)
+            {
+                string payload = this.ExtractBatchAsPayload();
+                try { HttpLogSink.SendBeacon(this.endpoint, payload); }
+                catch { /* fire-and-forget transport */ }
+            }
+        }
+
+        private void OnTimerTick()
```
gautam-achieveai, line 196: **[Gautam's review bot] [HIGH] `OnTimerTick` missing try/catch — unhandled exception in `setInterval` callback**

This is the callback for `timer.SetInterval`, which maps to `window.setInterval`. It runs outside any Logger or application-level try/catch. If `Flush()` throws — specifically if `ExtractBatchAsPayload()` → `LogJsonBuilder.BuildEnvelope()` throws during JSON serialization — the exception hits the browser's `setInterval` dispatcher as an **unhandled error**.

Since this fires on a periodic timer, this creates a **repeating** unhandled error every `flushIntervalMs` until the page is reloaded.

This is part of a **systematic gap**: `ExtractBatchAsPayload()` sits outside try/catch boundaries in all 4 call sites:
- `Flush()` line 108 — outside the inner try/catch
- `OnTimerTick()` here — no try/catch at all ← **worst case**
- `OnPageUnload()` line 166 — outside try/catch
- `Detach()` line 140 — outside try/catch (caught at Logger level)

`ExtractBatchAsPayload` is deceptively dangerous because it **destructively swaps the queue** before serialization. If the serialization then throws, the events are lost.

**Recommended systemic fix**: Widen try/catch in each site to cover `ExtractBatchAsPayload`, or make `Flush()` itself fully protected:
```csharp
private void OnTimerTick()
{
    if (this.detached) { return; }
    if (this.queue.Count > 0)
    {
        try { this.Flush(); }
        catch { /* timer callback must not throw */ }
    }
}
```
And in `Flush()` itself, merge the two try blocks:
```csharp
this.flushing = true;
try
{
    string payload = this.ExtractBatchAsPayload();
    HttpLogSink.PostPayload(this.endpoint, payload);
}
catch { /* payload construction or transport failure — fire-and-forget */ }
finally
{
    this.flushing = false;
}
```
Same pattern for `OnPageUnload()` and `Detach()`.

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        public void Handle(LogEvent evt)
+        {
+            if (this.detached) { return; }
+
+            // Overflow: drop oldest events, count the drops. We bias toward
+            // keeping the *most recent* entries because those are usually what
+            // the developer wants to see when triaging a crash.
+            while (this.queue.Count >= this.maxQueueSize)
```
gautam-achieveai: **[Gautam's review bot] [MEDIUM] O(n) `RemoveAt(0)` for overflow eviction**

`List<T>.RemoveAt(0)` compiles to `NativeArray.RemoveAt(0)` → `Array.splice(0, 1)` in JS. This is O(n) — the engine shifts all remaining elements left. With `maxQueueSize=500` under sustained overflow, each `Handle()` call triggers ~500 element shifts.

At typical sizes this is tolerable (~1-2ms/sec under burst), but it scales poorly with larger `maxQueueSize` values.

**Consider**: Replace with a single bulk-drop approach:
```csharp
if (this.queue.Count >= this.maxQueueSize)
{
    int excess = this.queue.Count - this.maxQueueSize + 1;
    this.droppedCount += excess;
    var trimmed = new List<LogEvent>();
    for (int i = excess; i < this.queue.Count; i++)
    {
        trimmed.Add(this.queue[i]);
    }
    this.queue = trimmed;
}
```
This does one O(n) pass instead of repeated splice operations.

Sources/Framework/Sunlight.Framework/Logging/NamedLogger.cs
```diff
@@ -0,0 +1,90 @@
...
+            Logger.DispatchInternal(LogLevel.Debug, this.category, message, null);
+        }
+
+        [Conditional("DEBUG")]
+        public void Debug(string message, string[] properties)
+        {
+            Logger.DispatchInternal(LogLevel.Debug, this.category, message, properties);
+        }
+
+        public void Info(string message)
```
gautam-achieveai, line 77: **[Gautam's review bot] [MEDIUM] `NamedLogger` methods skip `minLevel` check — inconsistent with static facade**

The static `Logger.Info/Warn/Error` methods all check `if (Logger.minLevel <= LogLevel.X)` before calling `DispatchInternal`. `NamedLogger` methods here call `DispatchInternal` unconditionally.

Functionally correct (DispatchInternal checks internally), but:

1. **Inconsistency** — consumers switching from `Logger.Info(...)` to `logger.Info(...)` unknowingly pay a function-call overhead for every filtered event.
2. **Misleading doc comment** — `DispatchInternal`'s XML doc says *""the facade and NamedLogger already apply the MinLevel check""* (Logger.cs:219-222). NamedLogger does **not**.

**Fix**: Add `minLevel` check to match the static pattern:
```csharp
public void Info(string message)
{
    if (Logger.MinLevel <= LogLevel.Info)
    {
        Logger.DispatchInternal(LogLevel.Info, this.category, message, null);
    }
}
```
And update the `DispatchInternal` doc comment to accurately describe the contract.

Test/Framework/Sunlight.Framework.Test/LoggerTests.cs
```diff
@@ -0,0 +1,379 @@
...
+            // indirectly: a third event triggers a flush and the queue resets.
+            // After flush we add a fourth event: the queue has one item, not four.
+            Logger.Info("three");
+            Logger.Info("four");
+
+            // Trigger timer tick — if the queue was flushed at size 3, only
+            // the fourth event remains and should flush now.
+            timer.Tick();
+
+            assert.IsTrue(true, "Batch size flush exercised without exceptions");
```
gautam-achieveai: **[Gautam's review bot] [MEDIUM] Test coverage gaps**

**Weak assertions**: `TestHttpLogSinkBatchesOnSize` (here) and `TestHttpLogSinkOverflowDropsOldest` (line 374) both end with `assert.IsTrue(true, ...)`. These prove no-throw but don't verify batch-size triggering or overflow behavior.

**Consider** adding a transport hook (`Action<string, string>`) to `HttpLogSink` constructor so tests can capture posted payloads and verify batch contents.

**LogJsonBuilder**: Zero test coverage. It's `internal static` and there's no `[InternalsVisibleTo]`. JSON shape is a wire-format contract with the server — field names, property serialization, `LevelToString`, envelope format are all untested. Consider adding `[InternalsVisibleTo(""Sunlight.Framework.Test"")]` and testing:
- Basic event shape (field names `ts`, `level`, `msg`, `cat`, `props`)
- Properties serialization (flat `[k,v,k,v]` → `{k:v}`)
- Odd-length properties (trailing key silently dropped)
- `BuildEnvelope` with `dropped` count

**Re-entrancy guard**: `Logger.dispatching` is documented as preventing infinite recursion but no test exercises it. A `ReentrantSink` that calls `Logger.Info()` from `Handle()` would anchor this guard against removal.

Sources/Framework/Sunlight.Framework/Logging/LogJsonBuilder.cs
```diff
@@ -0,0 +1,140 @@
...
+                    sb.Append(":");
+                    sb.Append(LogJsonBuilder.JsonEscape(props[i + 1]));
+                    i = i + 2;
+                }
+                sb.Append("}");
+            }
+
+            // CallContext correlation block — emitted as a single group so the
+            // server can cleanly branch on its presence.
+            var ctx = evt.Context;
```
gautam-achieveai, line 110: **[Gautam's review bot] [LOW] Stamp coupling: LogJsonBuilder reaches into all CallContext fields**

`AppendEvent` directly accesses `ctx.ActionId`, `ctx.TraceId`, `ctx.SpanId`, `ctx.ParentSpanId`, `ctx.Depth`. If `CallContext` adds fields (e.g., `BaggageItems`, `SamplingFlags`), this serializer must be updated in lockstep.

Non-blocking for v1 (only one serializer), but consider moving correlation JSON emission to `CallContext` itself (`AppendCorrelationJson(StringBuilder sb)`) so the context owns its serialization contract. This would also keep a future WebSocket sink in sync automatically.

gautam-achieveai (commented):
## [Gautam's review bot] Deep Review — PR !24: Structured Client Logging

**Linked**: User Story #11 — Add extensible structured client logging to the NScript framework

### Overall Assessment

This is a well-architected refactoring of the monolithic `Logger` into a sink-based pipeline. The design decisions are sound and well-documented — `ILogSink` is a clean interface, the `[Conditional("DEBUG")]` placement is correct, the NScript-specific constraints (minification-safe `string[]` properties, avoiding `[Script]` for-loop limitations, bypassing `XMLHttpRequest.OnBeforeSend` to prevent recursion) are all handled thoughtfully.

**However, there are two HIGH-severity exception-handling gaps that need fixing before merge.** Both can cause unhandled exceptions escaping the logging framework — which is particularly dangerous because `Logger.Error()` is called from `TaskScheduler`'s catch blocks.

### Review Agents Used
- Architecture review — clean, no layer violations or SOLID blockers
- Performance review — two MEDIUM findings (O(n) overflow, minLevel check)
- Exception handling review — two HIGH findings (escape paths)
- Test coverage review — MEDIUM gaps (LogJsonBuilder zero coverage, weak HttpLogSink assertions)

### Findings (6 total: 2 HIGH, 3 MEDIUM, 1 LOW)

See inline comments for details. Summary:

| # | Severity | Finding |
|---|----------|---------|
| 1 | **HIGH** | `DispatchInternal` outer try has no catch — exceptions escape `Logger.Error()` to crash TaskScheduler |
| 2 | **HIGH** | `ExtractBatchAsPayload()` outside try/catch in all 4 HttpLogSink call sites — `OnTimerTick` worst (repeating unhandled error in `setInterval`) |
| 3 | MEDIUM | `RemoveAt(0)` for overflow is O(n) per drop — `Array.splice(0,1)` in JS |
| 4 | MEDIUM | `NamedLogger` methods skip `minLevel` check (inconsistent with static facade) + misleading doc comment |
| 5 | MEDIUM | LogJsonBuilder has zero test coverage; HttpLogSink batch/overflow tests assert `true` (no behavioral verification) |
| 6 | LOW | Stamp coupling `LogJsonBuilder` to `CallContext`; namespace inconsistency vs `Binders`/`Observables` pattern |

### Strengths Worth Preserving
- `ILogSink` (Handle/Flush/Detach) — focused, cohesive lifecycle contract
- Re-entrancy guards at both levels (`Logger.dispatching` + `HttpLogSink.flushing`)
- `[Conditional]` kept off `ILogger` — prevents silent stripping bypass
- `LogEvent` immutability with `readonly` fields
- `IWindowTimer` injection enabling deterministic `ControllableTimer` tests
- `PostPayload` using raw XHR to avoid `OnBeforeSend` -> `Logger.*` recursion
- Lazy default `ConsoleSink` with `userConfigured` flag — clean backward compat

Test/Compiler/NScript.CLR.Test/LoggerStructureTests.cs
```diff
@@ -78,16 +82,44 @@ public void Logger_MinLevelPropertyExists()
...
+
+        [TestMethod]
+        public void Logger_LogLevelMethodsHavePropertiesOverload()
+        {
+            // WI-11 adds a (string, string[]) overload on each level so callers
+            // can attach structured key/value pairs without boxing. string[] is
+            // used (rather than object) to survive NScript minification.
+            var loggerType = GetLoggerType();
+
+            foreach (var name in new[] { "Debug", "Info", "Warn", "Error" })
```
gautam-achieveai: **[Gautam's review bot] [LOW] `Trace` omitted from both structure tests**

WI-11 added `Logger.Trace(string)` and `Logger.Trace(string, string[])` as new methods (lines 129-143 in Logger.cs). Both `Logger_LogLevelMethodsAcceptSingleStringParameter` and this new test only check Debug/Info/Warn/Error - `Trace` is absent from all three structure tests.

`Debug` is also `[Conditional("DEBUG")]` yet IS included in the loop, so the omission is not a deliberate skip-conditional-methods pattern. `[Conditional]` strips call sites, not the method definition - `Trace` is present in the Release DLL.

If someone accidentally removes the `Trace(string, string[])` overload, no structure test would catch it.

**Suggestion**: Add `Trace` to the level arrays in all three tests, or add a comment explaining why `Trace` is deliberately excluded.

gautam-achieveai (commented):
## [Gautam's review bot] Incremental Review - Iteration 2 (commit 1d56c47)

**Scope**: `LoggerStructureTests.cs` - 1 file changed (+44 / -12)

### Iteration 2 Changes - Assessment

The structural test updates are **clean and correct**:

1. **`Logger_EmitMethodHasScriptAttribute` renamed to `Logger_GetIsoTimestampHasScriptAttribute`** - Properly reflects the post-WI-11 method rename. Good comments explaining why `GetIsoTimestamp` is the surviving `[Script]`-bodied helper.

2. **`Logger_LogLevelMethodsAcceptSingleStringParameter`** - Tightened to specifically find the single-string overload via LINQ predicate (necessary now that `(string, string[])` overloads exist). Comment clearly states the back-compat guarantee intent.

3. **`Logger_LogLevelMethodsHavePropertiesOverload`** (new) - Validates the new `(string, string[])` overloads. Comment explains the `string[]` choice for minification safety. Well-structured, follows existing test patterns.

One low-severity observation posted inline below.

### Prior Findings Status - All 6 Open

**None of the 6 findings from iteration 1 have been addressed in this commit.** The production code (`Logger.cs`, `HttpLogSink.cs`, `NamedLogger.cs`, `LoggerTests.cs`, `LogJsonBuilder.cs`) is unchanged.

| # | Severity | Status | Finding |
|---|----------|--------|---------|
| 1 | **HIGH** | Open | `DispatchInternal` outer try has no `catch` - exceptions escape `Logger.Error()` |
| 2 | **HIGH** | Open | `ExtractBatchAsPayload()` outside try/catch in all 4 HttpLogSink call sites |
| 3 | MEDIUM | Open | `RemoveAt(0)` O(n) overflow eviction |
| 4 | MEDIUM | Open | `NamedLogger` skips `minLevel` check + misleading doc |
| 5 | MEDIUM | Open | LogJsonBuilder zero test coverage; HttpLogSink tests assert nothing |
| 6 | LOW | Open | Stamp coupling LogJsonBuilder to CallContext |

### Verdict: REQUEST CHANGES (maintained)

The iteration 2 test changes are fine. The two HIGH exception-handling findings from iteration 1 still need to be addressed before merge.

Sources/Framework/Sunlight.Framework/Logger.cs
gautam-achieveai: Resolved in 6fc9dd1. Added outer `catch` around the event-construction + fan-out block in `DispatchInternal` (Logger.cs:266-274). Exceptions from `EnsureSinkList`, `new ConsoleSink()`, `GetIsoTimestamp()`, or `new LogEvent(...)` are now swallowed so `Logger.Error()` stays safe to call from `TaskScheduler` catch blocks.

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
gautam-achieveai: Resolved in 6fc9dd1. Applied the recommended systemic fix: `Flush()` now wraps `ExtractBatchAsPayload` + transport in a single try/catch; `OnTimerTick`, `OnPageUnload`, and `Detach` all widened to include the payload build. Verified no call site extracts the batch outside a catch boundary.

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
gautam-achieveai: Resolved in 6fc9dd1. Replaced the `while (queue.Count >= maxQueueSize) RemoveAt(0)` loop with a single bulk trim: compute excess, increment `droppedCount` once, rebuild the tail into a new `List<LogEvent>`. One O(n) pass instead of O(n²) under sustained overflow.

Sources/Framework/Sunlight.Framework/Logging/NamedLogger.cs
gautam-achieveai: Resolved in 6fc9dd1. Every `NamedLogger` level method now guards `DispatchInternal` behind a `Logger.MinLevel <= LogLevel.X` check, matching the static facade. Updated the `DispatchInternal` doc comment to accurately describe the contract (NamedLogger must apply the gate itself; the internal check is a defensive backstop).

Test/Framework/Sunlight.Framework.Test/LoggerTests.cs
gautam-achieveai: Resolved in 6fc9dd1. Added an internal test-only `HttpLogSink` constructor that accepts a transport override lambda; the batch-size and overflow tests now capture serialized envelopes and assert on batch contents, dropped counts, and retained/dropped messages. Added `TestReentrancyGuardDropsNestedDispatch` to anchor the `Logger.dispatching` guard. Granted `InternalsVisibleTo("Sunlight.Framework.Test")` (via new `AssemblyInfo.cs") and added 6 direct `LogJsonBuilder` tests covering event shape, level mapping, flat-properties flattening, odd-length trailing-key drop, envelope with dropped count, and CallContext correlation.

Sources/Framework/Sunlight.Framework/Logging/LogJsonBuilder.cs
gautam-achieveai: Resolved in 6fc9dd1. Moved correlation serialization ownership onto `CallContext` itself: new `AppendCorrelationJson(StringBuilder, Func<string,string> escape)` emits the `actionId`/`traceId`/`spanId`/`parentSpanId`/`depth` block. `LogJsonBuilder` now delegates to it, passing `JsonEscape` as a lambda. Adding a future correlation field becomes a one-place change; a future WebSocket sink stays in sync automatically.

Test/Compiler/NScript.CLR.Test/LoggerStructureTests.cs
gautam-achieveai: Resolved in 6fc9dd1. Added `Trace` to both `Logger_LogLevelMethodsAcceptSingleStringParameter` and `Logger_LogLevelMethodsHavePropertiesOverload` arrays, with a comment in each noting why `[Conditional]` methods still need structural protection ([Conditional] strips call sites, not the method definition — the overload is still in the Release DLL).

gautam-achieveai:
Iteration 3 (commit 6fc9dd1) — addresses all 6 findings from the iteration-1 review plus the iteration-2 LOW finding.

**HIGH (2/2 resolved)**
- Logger.DispatchInternal: outer `catch` added for event-construction failures
- HttpLogSink: `ExtractBatchAsPayload` now inside try/catch in Flush / OnTimerTick / OnPageUnload / Detach

**MEDIUM (3/3 resolved)**
- HttpLogSink overflow: single-pass bulk trim replaces the O(n²) RemoveAt(0) loop
- NamedLogger: MinLevel check applied before DispatchInternal on every level; doc corrected
- Test coverage: transport override ctor for observable HttpLogSink tests; InternalsVisibleTo + 6 LogJsonBuilder tests; reentrancy guard test

**LOW (2/2 resolved)**
- Correlation serialization moved to CallContext.AppendCorrelationJson (removes stamp coupling)
- Trace added to both LoggerStructureTests overload arrays

**Verification**
- Release and Debug builds clean
- NScript.Converter.Test: 401/401
- NScript.CLR.Test: 94/94
- Framework browser tests: 175/175 (+7 new LoggerTests, +6 new LogJsonBuilderTests, +1 reentrancy test)

gautam-achieveai (commented):
## [Gautam's review bot] Follow-up Deep Review — PR !24: Structured Client Logging

**Linked**: User Story #11 — Add extensible structured client logging to the NScript framework

### Re-review Status

No new commits since the previous review (HEAD is still `1d56c47`). **All 7 prior findings remain unaddressed** — 0 of 7 threads resolved, no developer replies.

This follow-up dispatched three fresh specialized agents (exception-handling, performance, test-coverage) for a deeper pass. They uncovered **5 additional findings** (1 HIGH, 4 MEDIUM) that the initial review missed.

### Prior Findings (unchanged)

| # | Severity | Status | Finding |
|---|----------|--------|---------|
| P1 | **HIGH** | Open | `DispatchInternal` outer try has no catch — exceptions escape `Logger.Error()` |
| P2 | **HIGH** | Open | `ExtractBatchAsPayload()` outside try/catch in all 4 HttpLogSink call sites |
| P3 | MEDIUM | Open | `RemoveAt(0)` O(n) overflow eviction |
| P4 | MEDIUM | Open | `NamedLogger` skips `minLevel` check + misleading doc |
| P5 | MEDIUM | Open | LogJsonBuilder zero coverage; HttpLogSink tests assert nothing |
| P6 | LOW | Open | Stamp coupling LogJsonBuilder to CallContext |
| P7 | LOW | Open | `Trace` omitted from structure tests |

### New Findings (this review)

| # | Severity | Finding |
|---|----------|---------|
| N1 | **HIGH** | `maxQueueSize=0` infinite loop freezing browser tab (NScript-specific C#/JS semantic gap) |
| N2 | MEDIUM | `ExtractBatchAsPayload` destructive queue swap before serialization — irrecoverable data loss *(deepens P2)* |
| N3 | MEDIUM | `Detach()` has no exception isolation between cleanup steps |
| N4 | MEDIUM | `LogJsonBuilder` null property key produces invalid JSON — poisons entire batch |
| N5 | MEDIUM | Key test coverage gaps: Detach-with-queued-events, NamedLogger `string[]` properties, `ForCategory` cache identity *(deepens P5)* |

### Review Agents Used (this pass)
- **Exception handling review** — identified N1, N2, N3 + confirmed P1/P2 still valid
- **Performance review** — independently confirmed N1; found ConsoleSink eager serialization (deferred as optimization, not a bug)
- **Test coverage review** — identified N5 gaps
- **Review grader** — calibrated severity assignments, de-escalated 2 candidate findings (zombie timer, AddSink ordering) as unreachable in browser JS runtime

### Combined Tally

| Severity | Count |
|----------|-------|
| HIGH | 3 (P1 + P2 + N1) |
| MEDIUM | 7 (P3 + P4 + P5 + N2 + N3 + N4 + N5) |
| LOW | 2 (P6 + P7) |

### Verdict: REQUEST CHANGES (maintained and strengthened)

Three HIGH-severity findings now block merge:
1. **P1**: Exception escape from `Logger.Error()` crashes TaskScheduler
2. **P2**: Unhandled exceptions in `setInterval` callback (repeating)
3. **N1**: Browser tab freeze from `maxQueueSize=0` misconfiguration (NScript-specific infinite loop)

All three affect the error-handling code path (`TaskScheduler` → `Logger.Error` → sink pipeline) that the logging framework is specifically designed to support safely.

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
```diff
@@ -0,0 +1,282 @@
...
+
+        public HttpLogSink(
+            string endpoint,
+            int batchSize,
+            int flushIntervalMs,
+            int maxQueueSize,
+            IWindowTimer timer)
+        {
+            if (endpoint == null) { throw new ArgumentNullException("endpoint"); }
+            if (timer == null) { throw new ArgumentNullException("timer"); }
```
gautam-achieveai, line 76: **[Gautam's review bot] [HIGH] Missing constructor validation on `maxQueueSize` — causes infinite loop freezing browser tab**

The constructor validates `endpoint` and `timer` for null but has **zero validation** on the three numeric parameters (`batchSize`, `flushIntervalMs`, `maxQueueSize`). The `maxQueueSize=0` case is catastrophic.

In `Handle()` (line 83):
```csharp
while (this.queue.Count >= this.maxQueueSize)  // 0 >= 0 → always true
{
    this.queue.RemoveAt(0);  // silent no-op on empty JS array!
    this.droppedCount++;
}
```

This exploits a **C#-to-JS semantic gap** in NScript: `NativeArray<T>.RemoveAt` (line 251 of NativeArray.cs) bounds-checks with `index > this.length` (strict `>`, not `>=`). So `RemoveAt(0)` on an empty array evaluates `0 > 0` → false → no exception → `splice(0,1)` is a JS no-op → `Count` stays 0 → loop condition remains true → **infinite loop**.

In standard .NET this would throw `ArgumentOutOfRangeException` and be caught by `DispatchInternal`'s `catch {}`. In NScript JS, the main thread hangs permanently.

The call chain: `TaskScheduler.ExecuteTask` (line 417) → `Logger.Error(...)` → `DispatchInternal` → `HttpLogSink.Handle` → infinite loop → **browser tab frozen**.

**Two-part fix:**

1. Add constructor validation:
```csharp
if (maxQueueSize < 1) { throw new ArgumentException("maxQueueSize must be >= 1"); }
if (batchSize < 1) { throw new ArgumentException("batchSize must be >= 1"); }
if (flushIntervalMs < 1) { throw new ArgumentException("flushIntervalMs must be >= 1"); }
```

2. Change `while` to `if` as defense-in-depth (line 83). `Handle` adds exactly one event per call and JS is single-threaded — no concurrent mutation. `while` is unnecessary and escalates a misconfiguration from "buggy" to "infinite loop."

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
```diff
@@ -0,0 +1,282 @@
...
+
+        /// <summary>
+        /// Atomically swap the queue + <c>droppedCount</c> out, reset both, and
+        /// return the JSON envelope. Shared by <see cref="Flush"/>,
+        /// <see cref="Detach"/>, and <see cref="OnPageUnload"/> so the three
+        /// paths stay in sync and cannot drift.
+        /// </summary>
+        private string ExtractBatchAsPayload()
+        {
+            var batch = this.queue;
```
gautam-achieveai, line 265: **[Gautam's review bot] [MEDIUM] `ExtractBatchAsPayload` destructive swap ordering — irrecoverable data loss on serialization failure** *(deepens prior finding on line 146)*

This method destructively swaps the queue out **before** serialization:
```csharp
var batch = this.queue;               // reference to current queue
int dropped = this.droppedCount;
this.queue = new List<LogEvent>();      // ← queue replaced with empty list
this.droppedCount = 0;                 // ← counter zeroed
return LogJsonBuilder.BuildEnvelope(batch, dropped);  // ← if this throws, events gone
```

If `BuildEnvelope` throws (e.g., due to the null-property-key issue I flagged on `LogJsonBuilder.cs:90`, or any future serialization bug), the events are **irrecoverably lost** — `this.queue` already points to a fresh empty list, `batch` is a local that goes out of scope, and `droppedCount` was reset to 0 so the loss isn't even counted.

The prior review's recommendation to widen try/catch around `ExtractBatchAsPayload` is **necessary but insufficient** — it prevents exceptions from escaping but doesn't prevent data loss.

**Fix**: Serialize first, swap only on success:
```csharp
private string ExtractBatchAsPayload()
{
    var batch = this.queue;
    int dropped = this.droppedCount;
    string payload = LogJsonBuilder.BuildEnvelope(batch, dropped);
    // Swap only after successful serialization
    this.queue = new List<LogEvent>();
    this.droppedCount = 0;
    return payload;
}
```

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
```diff
@@ -0,0 +1,282 @@
...
+                this.flushing = false;
+            }
+        }
+
+        public void Detach()
+        {
+            if (this.detached) { return; }
+            this.detached = true;
+
+            if (this.timerHandle >= 0)
```
gautam-achieveai, line 171: **[Gautam's review bot] [MEDIUM] `Detach()` has no exception isolation between cleanup steps — one failure skips remaining cleanup**

The three cleanup steps are sequential with no independent exception protection:

1. `this.timer.ClearInterval(this.timerHandle)` — line 129
2. `this.RemoveUnloadHandler(this.unloadHandler)` — line 133
3. Final flush via `ExtractBatchAsPayload` + `SendBeacon` — lines 138-143

If step 1 throws, then `beforeunload`/`pagehide` event listeners **leak** (step 2 skipped) and buffered events are **silently lost** (step 3 skipped). `this.detached` is already `true`, so retrying `Detach()` is a no-op — **no recovery path**.

Distinct from the prior finding about `ExtractBatchAsPayload` outside try/catch — here the issue is any step failing cascades to skip all subsequent steps.

**Fix**: Isolate each step, matching the pattern already used in `Logger.ClearSinks`:
```csharp
public void Detach()
{
    if (this.detached) { return; }
    this.detached = true;

    if (this.timerHandle >= 0)
    {
        try { this.timer.ClearInterval(this.timerHandle); } catch { }
        this.timerHandle = -1;
    }

    try { this.RemoveUnloadHandler(this.unloadHandler); } catch { }
    this.unloadHandler = null;

    if (this.queue.Count > 0)
    {
        try
        {
            string payload = this.ExtractBatchAsPayload();
            HttpLogSink.SendBeacon(this.endpoint, payload);
        }
        catch { /* best-effort final flush */ }
    }
}
```

Sources/Framework/Sunlight.Framework/Logging/LogJsonBuilder.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                sb.Append(",\"props\":{");
+                bool first = true;
+                int n = props.Length;
+                int i = 0;
+                while (i + 1 < n)
+                {
+                    if (!first) { sb.Append(","); }
+                    first = false;
+                    sb.Append(LogJsonBuilder.JsonEscape(props[i]));
```
gautam-achieveai: **[Gautam's review bot] [MEDIUM] Null property key produces invalid JSON — poisons entire HttpLogSink batch**

`JsonEscape` delegates to `JSON.stringify(s)` (line 137). Per the doc comment and JS semantics, `JSON.stringify(null)` returns the bare four characters `null` (no quotes).

If a caller passes a null key — e.g., `new string[] { someVar, "value" }` where `someVar` is null — the output is:
```json
{"props":{null:"value"}}
```

Bare `null` as an object key is **invalid JSON** per RFC 8259 (keys must be quoted strings). Server-side parsers (`System.Text.Json`, `Newtonsoft.Json`, the `JavaScriptSerializer` used in `ClientLogs.ashx`) will **reject the entire payload**.

Since `HttpLogSink.ExtractBatchAsPayload` serializes an entire batch into one JSON envelope, one null property key from any event poisons **all** events in that batch.

**Fix**: Skip pairs with null keys:
```csharp
while (i + 1 < n)
{
    string key = props[i];
    if (key == null) { i = i + 2; continue; }
    if (!first) { sb.Append(","); }
    first = false;
    sb.Append(LogJsonBuilder.JsonEscape(key));
    sb.Append(":");
    sb.Append(LogJsonBuilder.JsonEscape(props[i + 1]));
    i = i + 2;
}
```

Test/Framework/Sunlight.Framework.Test/LoggerTests.cs
```diff
@@ -0,0 +1,379 @@
...
+    using System.Collections.Generic;
+    using SunlightUnit;
+    using Sunlight.Framework;
+
+    /// <summary>
+    /// QUnit tests for the structured logging pipeline added in WI-11.
+    /// Each test resets <see cref="Logger"/> state at start via
+    /// <see cref="ResetLogger"/> because the facade is a process-wide singleton.
+    /// </summary>
+    [TestFixture]
```
gautam-achieveai, line 19: **[Gautam's review bot] [MEDIUM] Additional test coverage gaps** *(extends prior finding at line 336)*

Beyond the LogJsonBuilder and weak-assertion gaps flagged previously, three high-value paths have zero test coverage:

**1. HttpLogSink Detach-with-queued-events** — `TestHttpLogSinkTimerClearedOnDetach` (line 342) removes the sink **without logging any events**. The final-flush branch (`HttpLogSink.cs:138-143`) — the **data-loss-prevention path** — is never exercised. This is the most consequential untested path in HttpLogSink.

**2. NamedLogger `string[]` properties overloads** — All NamedLogger tests call only single-string overloads. The properties test (`TestPropertiesPreservedThroughDispatch`, line 210) uses the **static** facade, not a NamedLogger. These `(string, string[])` overloads are the core **structured** part of "structured logging" — the headline feature of this PR — yet have zero test coverage through the category logger API.

**3. `ForCategory(null)` and `ForCategory("")` cache identity** — Tests verify null normalizes to empty and same-string returns same instance, but no test verifies that `ForCategory(null)` and `ForCategory("")` return the **same cached instance** — the semantic guarantee that null-normalization provides.

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
```diff
@@ -0,0 +1,352 @@
...
+        internal HttpLogSink(
+            string endpoint,
+            int batchSize,
+            int flushIntervalMs,
+            int maxQueueSize,
+            IWindowTimer timer,
+            Action<string, string> transportOverride)
+        {
+            if (endpoint == null) { throw new ArgumentNullException("endpoint"); }
+            if (timer == null) { throw new ArgumentNullException("timer"); }
```
gautam-achieveai, line 76: **[Gautam's review bot] [MEDIUM - non-blocking] Constructor still lacks numeric parameter validation**

*Follows up on N1 — the infinite loop (HIGH) is resolved by `while`→`if`, but this gap remains.*

The constructor validates `endpoint` and `timer` for null but accepts any integer for `batchSize`, `flushIntervalMs`, `maxQueueSize`. With `maxQueueSize=0`, overflow accounting inflates `droppedCount` (phantom drops). With `batchSize=0`, every `Handle` triggers a flush.

**Non-blocking**: constructor is `internal`, only framework team creates instances.

**Suggestion**:
```csharp
if (batchSize < 1) { throw new ArgumentOutOfRangeException("batchSize"); }
if (flushIntervalMs < 1) { throw new ArgumentOutOfRangeException("flushIntervalMs"); }
if (maxQueueSize < 1) { throw new ArgumentOutOfRangeException("maxQueueSize"); }
```

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
```diff
@@ -0,0 +1,352 @@
...
+        /// <see cref="Detach"/>, and <see cref="OnPageUnload"/> so the three
+        /// paths stay in sync and cannot drift.
+        /// </summary>
+        private string ExtractBatchAsPayload()
+        {
+            var batch = this.queue;
+            int dropped = this.droppedCount;
+            this.queue = new List<LogEvent>();
+            this.droppedCount = 0;
+            return LogJsonBuilder.BuildEnvelope(batch, dropped);
```
gautam-achieveai: **[Gautam's review bot] [MEDIUM - non-blocking] Swap-before-serialize still loses events on hypothetical failure**

*Follows up on N2.*

Queue replaced with a fresh list (line 251) *before* `BuildEnvelope` runs (line 253). If serialization throws, events are irrecoverably lost.

**Non-blocking**: `BuildEnvelope` is pure `StringBuilder.Append` + `JSON.stringify` on strings — effectively cannot throw in JS.

**Suggestion** (reorder 3 lines):
```csharp
private string ExtractBatchAsPayload()
{
    var batch = this.queue;
    int dropped = this.droppedCount;
    string result = LogJsonBuilder.BuildEnvelope(batch, dropped);
    this.queue = new List<LogEvent>();
    this.droppedCount = 0;
    return result;
}
```

Sources/Framework/Sunlight.Framework/Logging/LogJsonBuilder.cs
```diff
@@ -0,0 +1,133 @@
...
```
gautam-achieveai: **[Gautam's review bot] [MEDIUM - non-blocking] Null property key produces invalid JSON**

*Follows up on N4.*

`JSON.stringify(null)` returns bare `null` (no quotes) → invalid JSON key. Poisons entire batch.

Code already drops trailing keys from odd-length arrays but doesn't guard null keys.

**Suggestion**:
```csharp
while (i + 1 < n)
{
    string key = props[i];
    if (key == null) { i = i + 2; continue; }
    if (!first) { sb.Append(","); }
    first = false;
    sb.Append(LogJsonBuilder.JsonEscape(key));
    sb.Append(":");
    sb.Append(LogJsonBuilder.JsonEscape(props[i + 1]));
    i = i + 2;
}
```

gautam-achieveai (commented):
## [Gautam's review bot] Iteration 3 Review — PR !24: Structured Client Logging

**Linked**: User Story #11

### Verdict: ✅ APPROVE WITH SUGGESTIONS

All **blocking HIGH-severity findings are resolved**. The remaining items are defensive improvements — none represent production risk.

---

### Prior Findings (P1–P7): All Resolved ✅

| # | Severity | Status | Resolution |
|---|----------|--------|------------|
| P1 | **HIGH** | ✅ Resolved | Outer `catch` added in `DispatchInternal` (Logger.cs:266–274) |
| P2 | **HIGH** | ✅ Resolved | All 4 `ExtractBatchAsPayload` call sites wrapped in try/catch |
| P3 | MEDIUM | ✅ Resolved | `while/RemoveAt(0)` → single-pass bulk trim (O(n) vs O(n²)) |
| P4 | MEDIUM | ✅ Resolved | MinLevel gate on all 10 `NamedLogger` methods; doc updated |
| P5 | MEDIUM | ✅ Resolved | Transport override ctor, 6 LogJsonBuilder tests, reentrancy test |
| P6 | LOW | ✅ Resolved | `CallContext.AppendCorrelationJson` owns correlation serialization |
| P7 | LOW | ✅ Resolved | `Trace` added to both structure test arrays |

---

### Pass 3 Findings (N1–N5): Re-Evaluated

#### N1 — `maxQueueSize=0` → Downgraded HIGH → MEDIUM (non-blocking)
The `while` → `if` change eliminates the infinite loop. Remaining gap: no constructor validation on numeric params.

#### N2 — `ExtractBatchAsPayload` destructive swap → MEDIUM (non-blocking)
Queue swapped before serialization. `BuildEnvelope` is pure string concatenation — cannot throw in JS. Fix: reorder 3 lines.

#### N3 — `Detach()` cleanup isolation → Downgraded MEDIUM → LOW
`clearInterval()` and `removeEventListener()` spec-guaranteed not to throw.

#### N4 — Null property key → invalid JSON → MEDIUM (non-blocking)
Requires caller bug. Code already handles odd-length arrays but not null keys.

#### N5 — Test coverage gaps → MEDIUM/LOW (non-blocking)
Detach-with-queued-events, NamedLogger `string[]` props, ForCategory cache identity.

---

### Non-Blocking Suggestions
1. Add `< 1` guards on numeric ctor params (~3 lines)
2. Reorder `ExtractBatchAsPayload` to serialize-then-swap (~3 lines)
3. Skip null property keys in `LogJsonBuilder` (~2 lines)
4. Add test for Detach-with-queued-events (~10 lines)
5. Add test for `namedLogger.Info(msg, props)` (~8 lines)

See inline comments for details.

### Review Agents Used (Pass 4)
- Exception handling, Performance, Test coverage, Review grader

### Combined Tally
| Severity | Resolved | Open (non-blocking) |
|----------|----------|---------------------|
| HIGH | 3 | 0 |
| MEDIUM | 3 | 4 |
| LOW | 2 | 2 |

**All HIGHs resolved. PR is ready to merge.** 🚀

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
gautam-achieveai, line 76: **[Gautam's review bot] [MEDIUM - non-blocking] Constructor still lacks numeric parameter validation**

*Follows up on N1 — the infinite loop (HIGH) is resolved by `while`→`if`, but this gap remains.*

The constructor validates `endpoint` and `timer` for null but accepts any integer for `batchSize`, `flushIntervalMs`, `maxQueueSize`. With `maxQueueSize=0`, overflow accounting inflates `droppedCount` (phantom drops). With `batchSize=0`, every `Handle` triggers a flush.

**Non-blocking**: constructor is `internal`, only framework team creates instances.

**Suggestion**:
```csharp
if (batchSize < 1) { throw new ArgumentOutOfRangeException("batchSize"); }
if (flushIntervalMs < 1) { throw new ArgumentOutOfRangeException("flushIntervalMs"); }
if (maxQueueSize < 1) { throw new ArgumentOutOfRangeException("maxQueueSize"); }
```

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
gautam-achieveai: **[Gautam's review bot] [MEDIUM - non-blocking] Swap-before-serialize still loses events on hypothetical failure**

*Follows up on N2.*

Queue replaced with a fresh list (line 251) *before* `BuildEnvelope` runs (line 253). If serialization throws, events are irrecoverably lost.

**Non-blocking**: `BuildEnvelope` is pure `StringBuilder.Append` + `JSON.stringify` on strings — effectively cannot throw in JS.

**Suggestion** (reorder 3 lines):
```csharp
private string ExtractBatchAsPayload()
{
    var batch = this.queue;
    int dropped = this.droppedCount;
    string result = LogJsonBuilder.BuildEnvelope(batch, dropped);
    this.queue = new List<LogEvent>();
    this.droppedCount = 0;
    return result;
}
```

Sources/Framework/Sunlight.Framework/Logging/LogJsonBuilder.cs
gautam-achieveai: **[Gautam's review bot] [MEDIUM - non-blocking] Null property key produces invalid JSON**

*Follows up on N4.*

`JSON.stringify(null)` returns bare `null` (no quotes) → invalid JSON key. Poisons entire batch.

Code already drops trailing keys from odd-length arrays but doesn't guard null keys.

**Suggestion**:
```csharp
while (i + 1 < n)
{
    string key = props[i];
    if (key == null) { i = i + 2; continue; }
    if (!first) { sb.Append(","); }
    first = false;
    sb.Append(LogJsonBuilder.JsonEscape(key));
    sb.Append(":");
    sb.Append(LogJsonBuilder.JsonEscape(props[i + 1]));
    i = i + 2;
}
```

gautam-achieveai (commented):
## [Gautam's review bot] Iteration 3 Review — PR !24: Structured Client Logging

**Linked**: User Story #11

### Verdict: ✅ APPROVE WITH SUGGESTIONS

All **blocking HIGH-severity findings are resolved**. The remaining items are defensive improvements — none represent production risk.

---

### Prior Findings (P1–P7): All Resolved ✅

| # | Severity | Status | Resolution |
|---|----------|--------|------------|
| P1 | **HIGH** | ✅ Resolved | Outer `catch` added in `DispatchInternal` (Logger.cs:266–274) |
| P2 | **HIGH** | ✅ Resolved | All 4 `ExtractBatchAsPayload` call sites wrapped in try/catch |
| P3 | MEDIUM | ✅ Resolved | `while/RemoveAt(0)` → single-pass bulk trim (O(n) vs O(n²)) |
| P4 | MEDIUM | ✅ Resolved | MinLevel gate on all 10 `NamedLogger` methods; doc updated |
| P5 | MEDIUM | ✅ Resolved | Transport override ctor, 6 LogJsonBuilder tests, reentrancy test |
| P6 | LOW | ✅ Resolved | `CallContext.AppendCorrelationJson` owns correlation serialization |
| P7 | LOW | ✅ Resolved | `Trace` added to both structure test arrays |

---

### Pass 3 Findings (N1–N5): Re-Evaluated

#### N1 — `maxQueueSize=0` → Downgraded HIGH → MEDIUM (non-blocking)
The `while` → `if` change eliminates the infinite loop. Remaining gap: no constructor validation on numeric params.

#### N2 — `ExtractBatchAsPayload` destructive swap → MEDIUM (non-blocking)
Queue swapped before serialization. `BuildEnvelope` is pure string concatenation — cannot throw in JS. Fix: reorder 3 lines.

#### N3 — `Detach()` cleanup isolation → Downgraded MEDIUM → LOW
`clearInterval()` and `removeEventListener()` spec-guaranteed not to throw.

#### N4 — Null property key → invalid JSON → MEDIUM (non-blocking)
Requires caller bug. Code already handles odd-length arrays but not null keys.

#### N5 — Test coverage gaps → MEDIUM/LOW (non-blocking)
Detach-with-queued-events, NamedLogger `string[]` props, ForCategory cache identity.

---

### Non-Blocking Suggestions
1. Add `< 1` guards on numeric ctor params (~3 lines)
2. Reorder `ExtractBatchAsPayload` to serialize-then-swap (~3 lines)
3. Skip null property keys in `LogJsonBuilder` (~2 lines)
4. Add test for Detach-with-queued-events (~10 lines)
5. Add test for `namedLogger.Info(msg, props)` (~8 lines)

See inline comments for details.

### Review Agents Used (Pass 4)
- Exception handling, Performance, Test coverage, Review grader

### Combined Tally
| Severity | Resolved | Open (non-blocking) |
|----------|----------|---------------------|
| HIGH | 3 | 0 |
| MEDIUM | 3 | 4 |
| LOW | 2 | 2 |

**All HIGHs resolved. PR is ready to merge.** 🚀

Sources/Framework/Sunlight.Framework/Logger.cs
gautam-achieveai: **[Gautam's review bot]** Verified in 6fc9dd1. Outer catch correctly covers event construction + lazy ConsoleSink allocation. The finally block properly resets `dispatching`. **P1 resolved.**

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
gautam-achieveai: **[Gautam's review bot]** Verified in 6fc9dd1. All 4 call sites now protected: `Flush` (single try/catch), `OnTimerTick` (wraps Flush), `OnPageUnload` (wraps payload+transport), `Detach` (wraps payload+transport). No call site extracts the batch outside a catch boundary. **P2 resolved.**

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
gautam-achieveai: **[Gautam's review bot]** Verified in 6fc9dd1. Single-pass bulk trim is correct O(n). The `while` to `if` change is provably safe for single-threaded JS. **P3 resolved.**

Sources/Framework/Sunlight.Framework/Logging/NamedLogger.cs
gautam-achieveai: **[Gautam's review bot]** Verified in 6fc9dd1. All 10 NamedLogger methods now have MinLevel guard. Doc comment accurately describes the contract. **P4 resolved.**

Test/Framework/Sunlight.Framework.Test/LoggerTests.cs
gautam-achieveai: **[Gautam's review bot]** Verified in 6fc9dd1. Transport override ctor enables behavioral assertions. 6 LogJsonBuilder tests cover the wire-format contract. Reentrancy guard test anchors the `dispatching` flag. **P5 resolved.** (N5 test gaps for Detach-with-events, NamedLogger props, and ForCategory cache identity remain as non-blocking suggestions.)

Sources/Framework/Sunlight.Framework/Logging/LogJsonBuilder.cs
gautam-achieveai: **[Gautam's review bot]** Verified in 6fc9dd1. `CallContext.AppendCorrelationJson` correctly owns the serialization contract. Lambda wrapper avoids the NScript method-group delegate limitation. **P6 resolved.**

Test/Compiler/NScript.CLR.Test/LoggerStructureTests.cs
gautam-achieveai: **[Gautam's review bot]** Verified in 6fc9dd1. Trace added to both arrays with clear comments explaining why [Conditional] methods need structural protection. **P7 resolved.**

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
gautam-achieveai: **[Gautam's review bot]** Re-evaluated in Pass 4. The `while` to `if` change in iteration 3 **definitively eliminates the infinite loop** — downgrading from HIGH to **MEDIUM (non-blocking)**.

Remaining gap: no constructor validation on numeric params. With `maxQueueSize=0`, overflow bookkeeping inflates `droppedCount` (phantom drops). Since the constructor is `internal` and documented examples use sensible values, this is a defensive improvement, not a blocker. See inline comment on line 76 for the suggested fix.

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
gautam-achieveai: **[Gautam's review bot]** Re-evaluated in Pass 4. Still structurally present but confirmed as **MEDIUM (non-blocking)**.

`BuildEnvelope` is pure `StringBuilder.Append` + `JSON.stringify` on strings — effectively cannot throw in the JS runtime. JS is single-threaded so no concurrent mutation risk. The fix (serialize-then-swap) is trivial — just reorder 3 lines. See inline comment on line 253 for the suggested fix.

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
gautam-achieveai: **[Gautam's review bot]** Re-evaluated in Pass 4. Downgrading from MEDIUM to **LOW**.

Both `clearInterval()` and `removeEventListener()` are **spec-guaranteed not to throw** per WHATWG/W3C. The `ControllableTimer` test double also does not throw. Risk is purely theoretical. The final-flush step (lines 177-183) is already protected with try/catch.

Sources/Framework/Sunlight.Framework/Logging/LogJsonBuilder.cs
gautam-achieveai: **[Gautam's review bot]** Re-evaluated in Pass 4. Confirmed as **MEDIUM (non-blocking)**.

Requires a caller bug (null in `string[]`). The code already defensively handles odd-length arrays — adding a null-key guard would be consistent. See inline comment on line 90 for the suggested fix.

Test/Framework/Sunlight.Framework.Test/LoggerTests.cs
gautam-achieveai: **[Gautam's review bot]** Re-evaluated in Pass 4. Three gaps confirmed as **non-blocking**:

1. **Detach-with-queued-events** (MEDIUM) — final flush path never exercised. The transport-override ctor makes this easy (~10 lines).
2. **NamedLogger `string[]` properties** (MEDIUM) — headline structured feature untested through category API.
3. **ForCategory(null)/ForCategory("")** cache identity (LOW) — normalization tested, identity not.

All are good suggestions for a follow-up but none block this PR.

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
gautam-achieveai: Fixed in commit 0b22505a. Added fail-fast guards in the internal ctor for `batchSize`, `flushIntervalMs`, `maxQueueSize` (all must be `>= 1`). NScript's mscorlib does not expose `ArgumentOutOfRangeException`, so the guards throw `ArgumentException` with the parameter name in the message. Coverage: the existing test suite exercises only valid configurations; invalid-argument tests weren't added to avoid bloating the suite for a one-line ctor guard.

gautam-achieveai: Fixed in commit 0b22505a. Added fail-fast guards in the internal ctor for `batchSize`, `flushIntervalMs`, `maxQueueSize` (all must be `>= 1`). NScript's mscorlib does not expose `ArgumentOutOfRangeException`, so the guards throw `ArgumentException` with the parameter name in the message. Coverage: the existing test suite exercises only valid configurations; invalid-argument tests weren't added to avoid bloating the suite for a one-line ctor guard.

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
gautam-achieveai: Fixed in commit 0b22505a. `ExtractBatchAsPayload` now calls `LogJsonBuilder.BuildEnvelope` BEFORE resetting `queue`/`droppedCount`. On a serialization throw, the buffered events stay in place for the next flush attempt instead of being silently erased. Added a remark on the method documenting the serialize-then-swap invariant.

gautam-achieveai: Fixed in commit 0b22505a. `ExtractBatchAsPayload` now calls `LogJsonBuilder.BuildEnvelope` BEFORE resetting `queue`/`droppedCount`. On a serialization throw, the buffered events stay in place for the next flush attempt instead of being silently erased. Added a remark on the method documenting the serialize-then-swap invariant.

Sources/Framework/Sunlight.Framework/Logging/LogJsonBuilder.cs
gautam-achieveai: Fixed in commit 0b22505a. `LogJsonBuilder.AppendEvent` now skips property pairs whose key is `null` instead of emitting `"null":"value"` (which would corrupt the outer object). New test `TestLogJsonBuilderNullKeySkipped` anchors this behavior with a string[] containing a null slot.

gautam-achieveai: Fixed in commit 0b22505a. `LogJsonBuilder.AppendEvent` now skips property pairs whose key is `null` instead of emitting `"null":"value"` (which would corrupt the outer object). New test `TestLogJsonBuilderNullKeySkipped` anchors this behavior with a string[] containing a null slot.

Test/Framework/Sunlight.Framework.Test/LoggerTests.cs
gautam-achieveai: Three new tests added in commit 0b22505a:
- `TestHttpLogSinkDetachFlushesQueuedEvents` — residual events drain through the transport on sink removal.
- `TestNamedLoggerPassesPropertiesThrough` — properties array is forwarded to `DispatchInternal` unchanged on the category path.
- `TestForCategoryNullAndEmptyShareCacheSlot` — null normalizes to "" and hits the same cache bucket.

All framework browser tests green (179/179, +4 new including the N4 null-key test).

gautam-achieveai:
## Iter-4 update (commit 0b22505a)

All 5 non-blocking suggestions from the iter-3 "APPROVE WITH SUGGESTIONS" review addressed:

- **N1** — `HttpLogSink` ctor now rejects `batchSize`/`flushIntervalMs`/`maxQueueSize` below 1 via `ArgumentException`.
- **N2** — `ExtractBatchAsPayload` serializes the envelope BEFORE resetting the queue; buffered events survive any hypothetical `BuildEnvelope` throw.
- **N4** — `LogJsonBuilder.AppendEvent` skips property pairs with a null key instead of emitting `"null":"value"`.
- **N5** — Three new browser tests: `TestHttpLogSinkDetachFlushesQueuedEvents`, `TestNamedLoggerPassesPropertiesThrough`, `TestForCategoryNullAndEmptyShareCacheSlot`. Also `TestLogJsonBuilderNullKeySkipped` anchors N4.

Verification:
- Release + Debug builds clean.
- `NScript.Converter.Test` 401/401.
- `NScript.CLR.Test` 94/94.
- Framework browser tests 179/179 (+4 new).

CI status: one of two parallel GHA legs flaked on a "GeneratedScripts directory not found" path — identical to runs on earlier commits, unrelated to these edits. Triggered a re-run.

gautam-achieveai (commented):
**[Gautam's review bot]** Pass 5 — Iteration 4 incremental review (0b22505)

## ✅ APPROVE

All 4 non-blocking suggestions from Pass 4 addressed in this iteration:

| Finding | Status | Detail |
|---------|--------|--------|
| **N1** (MEDIUM) | ✅ Resolved | Constructor validation guards added — `batchSize`/`flushIntervalMs`/`maxQueueSize` reject `< 1` with `ArgumentException`. Fail-fast before any state mutation or resource allocation. |
| **N2** (MEDIUM) | ✅ Resolved | `ExtractBatchAsPayload` now serializes **before** resetting queue/droppedCount. A `BuildEnvelope` throw preserves events for the next flush attempt. |
| **N4** (MEDIUM) | ✅ Resolved | Null property keys silently skipped — prevents invalid JSON that would poison entire HttpLogSink batch. |
| **N5** (MEDIUM) | ✅ Resolved | All 3 test gaps filled: `TestHttpLogSinkDetachFlushesQueuedEvents`, `TestNamedLoggerPassesPropertiesThrough`, `TestForCategoryNullAndEmptyShareCacheSlot` + bonus `TestLogJsonBuilderNullKeySkipped`. |

**Remaining from prior passes:** N3 (Detach cleanup isolation) — LOW, non-blocking. `clearInterval`/`removeEventListener` are spec-guaranteed no-throw.

### New non-blocking observations (not blocking merge):

**F1 (MEDIUM):** Constructor validation guards have no unit tests. The guards prevent the three most dangerous misconfiguration modes, but since they're straightforward `< 1` checks on an `internal` API, risk of regression is low. Trivially addable as follow-up.

**F2 (LOW):** `ExtractBatchAsPayload` serialize-before-reset ordering isn't directly tested (hard to test without serialization failure seam). Well-documented in `<remarks>` block and implicitly tested by integration tests.

### Agents used
- `exception-handling-review`: 0 findings — all changes correct
- `test-coverage-review`: F1 (constructor tests), F2 (ordering test)
- `review-grader`: De-escalated F1 HIGH→MEDIUM, F2 MEDIUM→LOW; confirmed APPROVE

### Overall (5 passes, 4 iterations)
14 findings total across 5 review passes. 11 resolved, 1 LOW open (non-blocking), 2 new non-blocking observations. The author has been exceptionally responsive — every finding from every pass addressed. Architecture is strong and production-safe.

Linked: #11

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
```diff
@@ -0,0 +1,369 @@
...
+            // Numeric guards: zero/negative values would make the sink silently
+            // pathological — a zero batchSize flushes on every event, zero
+            // flushIntervalMs makes setInterval a tight loop, and zero
+            // maxQueueSize means Handle() always drops. Fail fast at ctor time
+            // instead of producing confusing runtime behavior. NScript's
+            // mscorlib does not expose ArgumentOutOfRangeException, so we use
+            // ArgumentException with a descriptive parameter name.
+            if (batchSize < 1) { throw new ArgumentException("batchSize must be >= 1"); }
+            if (flushIntervalMs < 1) { throw new ArgumentException("flushIntervalMs must be >= 1"); }
+            if (maxQueueSize < 1) { throw new ArgumentException("maxQueueSize must be >= 1"); }
```
gautam-achieveai, line 86: **[Gautam's review bot]** ✅ **N1 resolved.** Constructor validation guards correctly reject `< 1` values before any state mutation or resource allocation (`SetInterval`/`InstallUnloadHandler` at lines 103-104). Using `ArgumentException` instead of `ArgumentOutOfRangeException` is justified by NScript's mscorlib constraint — well-documented in the comment block.

📝 **Non-blocking (F1):** Consider adding unit tests for these guards as a follow-up — three `assert.Throws` calls using the existing `ControllableTimer` infrastructure would anchor the fail-fast contract.

Sources/Framework/Sunlight.Framework/Logging/HttpLogSink.cs
```diff
@@ -0,0 +1,369 @@
...
+        /// Serialization happens BEFORE the queue is reset so that a
+        /// <c>BuildEnvelope</c> throw leaves the buffered events in place for
+        /// the next flush attempt. Resetting first would silently drop an
+        /// entire batch on any JSON build failure.
+        /// </remarks>
+        private string ExtractBatchAsPayload()
+        {
+            var batch = this.queue;
+            int dropped = this.droppedCount;
+            string payload = LogJsonBuilder.BuildEnvelope(batch, dropped);
```
gautam-achieveai, line 267: **[Gautam's review bot]** ✅ **N2 resolved.** Serialization now happens before queue reset. If `BuildEnvelope` throws, `this.queue` and `this.droppedCount` are preserved — the `Flush()` caller's catch handles the exception and the next timer tick retries with events intact. The `<remarks>` block clearly documents the invariant.

Sources/Framework/Sunlight.Framework/Logging/LogJsonBuilder.cs
```diff
@@ -0,0 +1,138 @@
...
+            if (props != null && props.Length >= 2)
+            {
+                sb.Append(",\"props\":{");
+                bool first = true;
+                int n = props.Length;
+                int i = 0;
+                while (i + 1 < n)
+                {
+                    string key = props[i];
+                    if (key == null) { i = i + 2; continue; }
```
gautam-achieveai, line 92: **[Gautam's review bot]** ✅ **N4 resolved.** Null key skip correctly advances by 2 (`i = i + 2; continue;`) maintaining stride alignment. Consistent with the existing odd-trailing-key drop semantics. Good test coverage via `TestLogJsonBuilderNullKeySkipped`.

Test/Framework/Sunlight.Framework.Test/LoggerTests.cs
```diff
@@ -0,0 +1,648 @@
...
+                "m",
+                new string[] { "k1", "v1", null, "orphan-val", "k2", "v2" },
+                null);
+
+            string json = LogJsonBuilder.BuildEvent(evt);
+
+            assert.IsTrue(json.IndexOf("\"k1\":\"v1\"") >= 0, "First pair kept");
+            assert.IsTrue(json.IndexOf("\"k2\":\"v2\"") >= 0, "Third pair kept");
+            assert.IsTrue(json.IndexOf("\"null\"") < 0, "Null key is not serialized as the literal \"null\"");
+            assert.IsTrue(json.IndexOf("orphan-val") < 0, "Value paired with null key is dropped");
```
gautam-achieveai, line 562: **[Gautam's review bot]** ✅ **N5 resolved (null-key anchor test).** Excellent coverage: tests both kept pairs (k1/v1, k2/v2), absence of literal `"null"`, and dropped orphan value. Placing the null pair in the middle validates that iteration continues correctly past the skip.

Test/Framework/Sunlight.Framework.Test/LoggerTests.cs
```diff
@@ -0,0 +1,648 @@
...
+
+            Logger.Info("queued-1");
+            Logger.Info("queued-2");
+            assert.Equal(0, payloads.Count, "No flush yet — below batchSize");
+
+            Logger.RemoveSink(sink);
+
+            assert.Equal(1, payloads.Count, "Detach drained the queue through the transport");
+            assert.IsTrue(payloads[0].IndexOf("\"msg\":\"queued-1\"") >= 0, "Residual event 'queued-1' included");
+            assert.IsTrue(payloads[0].IndexOf("\"msg\":\"queued-2\"") >= 0, "Residual event 'queued-2' included");
```
gautam-achieveai, line 587: **[Gautam's review bot]** ✅ **N5 resolved (Detach-with-events).** This exercises the most consequential previously-untested path in HttpLogSink — the data-loss-prevention flush on Detach. Good structure: pre-condition assertion (no premature flush), post-condition on count and content verification.

Test/Framework/Sunlight.Framework.Test/LoggerTests.cs
```diff
@@ -0,0 +1,648 @@
...
+            ResetLogger();
+            var fake = new FakeSink();
+            Logger.AddSink(fake);
+
+            var props = new string[] { "userId", "42", "op", "save" };
+            var log = Logger.ForCategory("named-props");
+            log.Info("named-with-props", props);
+
+            assert.Equal(1, fake.Events.Count, "One event dispatched through NamedLogger");
+            assert.StrictEqual(props, fake.Events[0].Properties, "Properties array forwarded unchanged");
```
gautam-achieveai, line 606: **[Gautam's review bot]** ✅ **N5 resolved (NamedLogger properties).** `StrictEqual` for reference equality is the correct assertion — verifies the array is forwarded unchanged, not copied. This locks in the structured-logging contract through the category API.

Test/Framework/Sunlight.Framework.Test/LoggerTests.cs
```diff
@@ -0,0 +1,648 @@
...
+        {
+            // ForCategory(null) normalizes to the empty string and then looks
+            // up the same cache bucket as ForCategory(""). Without this
+            // guarantee, a caller that drifts between null and "" would get
+            // two distinct NamedLogger instances silently.
+            ResetLogger();
+            var nullLog = Logger.ForCategory(null);
+            var emptyLog = Logger.ForCategory(string.Empty);
+
+            assert.StrictEqual(emptyLog, nullLog,
```
gautam-achieveai, line 621: **[Gautam's review bot]** ✅ **N5 resolved (cache identity).** `StrictEqual` correctly verifies same cached instance, not just logical equality. Anchors the null→empty normalization guarantee.

gautam-achieveai:
Can you explain me how are you testing / validating that during release builds, we're not generating log lines for trace and debug logs?

gautam-achieveai:
[Gautam's Dev bot]

Good question — honest answer: **we test the *mechanism* that enables stripping, not the stripped JS output itself**. Here's the breakdown:

## The mechanism (Roslyn, not NScript)

`Logger.Trace` / `Logger.Debug` carry `[Conditional("DEBUG")]`. Roslyn's documented contract: at a caller's compilation time, if the caller's `DefineConstants` does **not** include `DEBUG`, Roslyn removes every call site — *including argument evaluation* — from the caller's bound tree.

The NScript-specific piece: stripping happens **pre-`OnBoundExpressionGenerated`**, which is the hook Stage 1 uses to capture the AST. So NScript's Stage 2 JS emitter never sees those calls, and the generated JavaScript contains no reference to them.

Two important nuances (noted in `Logger.cs:28-34` and in `LoggerStructureTests.cs:89-92` / `:113-116`):
- `[Conditional]` strips **call sites**, not **method definitions**. The `Trace`/`Debug` method bodies remain in the Release `Sunlight.Framework.dll` and still ship. Zero-size they are not — but they are unreferenced and get dead-code-eliminated in Stage 2 if no caller survives.
- Stripping is **per-assembly at the caller's compilation**, not the framework's. A `DEBUG`-built caller consuming a `Release` framework DLL *still keeps* its `Trace`/`Debug` calls.

## What the PR actually asserts

`Test/Compiler/NScript.CLR.Test/LoggerStructureTests.cs` opens the Release-built `NScriptToolSet/lib/Release/Sunlight.Framework.dll` via Cecil and asserts:
- `Logger_LogLevelMethodsAcceptSingleStringParameter` — `Trace(string)` / `Debug(string)` **definitions** still exist (guards against accidental removal).
- `Logger_LogLevelMethodsHavePropertiesOverload` — same for the `(string, string[])` overloads.

And `Test/Framework/.../LoggerTests.cs:285-312` proves the positive path: when the caller compiles under `DEBUG` (which the framework-test assembly does), `Logger.Trace(...)` does dispatch and produces an event. That confirms the method works — but confirms nothing about Release stripping.

## The gap

There is **no automated test in this PR that picks up a Release-compiled caller's emitted JS and asserts "no `Trace`/`Debug` call site remains"**. We lean on Roslyn's well-documented `[Conditional]` behavior — the same guarantee every .NET project relies on for `Debug.Assert`.

## If you want belt-and-suspenders coverage

Happy to add a test along these lines (say the word):
1. A small caller fixture (e.g. `TraceStrippingProbe.cs`) that calls `Logger.Trace("X")` and `Logger.Debug("Y")` alongside a sentinel `Logger.Info("SENTINEL")`.
2. Compile it twice — once with `DEBUG` in `DefineConstants`, once without.
3. Read the emitted JS (or the embedded `$$BstInfo$$` bound tree) and assert: `SENTINEL` appears in both outputs; `"X"` / `"Y"` appear **only** in the DEBUG output.

This would lock the guarantee into CI rather than deferring to Roslyn documentation. Want me to land it as a follow-up?

Issue #12, opened 2026-04-10
Complete source mapping support for C#, XWML, and Razor templates


gautam-achieveai:
## Summary
The NScript compiler has partial source mapping support (`Sources/Compiler/SourceMap/`, `JSWriter.cs`), but coverage is incomplete. Many JS constructs that represent C# code lack proper mappings (variable declarations, for-loop variables, etc.), templates (XWML and Razor) have no source map integration at all, variable name mappings are never populated, and the source file server is a legacy ASP.NET WebForms handler (`SrcMapper.ashx`).

This Feature tracks a phased effort to bring source mapping to production quality.

## Value
Developers debugging NScript-compiled JS in the browser need accurate mapping back to original C# and template sources. Complete source maps enable:
- Stepping through original C# code in browser DevTools
- Seeing original variable names instead of minified identifiers
- Debugging XWML and Razor template bindings
- Linking directly to source files in ADO/GitHub repos without local code access

**Priority:** Medium

## Current State (from codebase research)
- `SourceMap` class implements V3 source map format with Base64 VLQ encoding
- `JSWriter.cs` emits per-token mappings but only for tokens that carry `Location` data — many constructs are missing
- `names` array in source map is supported in the data model but never populated
- `SrcMapper.ashx` is an ASP.NET WebForms handler (not ASP.NET Core)
- **Zero tests** exist for source map correctness
- XWML (`XwmlParser/CodeGenerator.cs`) and Razor (`RazorSkinParser`) have no source map integration

## Sub-Issues
This effort is broken into the following sub-issues (will be linked after creation):
1. Fix incomplete source mappings for C# → JS code generation
2. Add source mapping support for XWML templates
3. Add source mapping support for Razor/Skin templates
4. Map original variable/symbol names in source maps
5. Add source map verification tooling and test suite
6. Replace SrcMapper.ashx with ASP.NET Core source file handler
7. Support linking source maps to ADO/GitHub repository files

## Phasing
- **Phase 1:** Verification tooling and test suite (sub-issue 5) — establish the testing foundation first
- **Phase 2:** Fix C# mapping gaps (sub-issue 1) — most impactful for day-to-day debugging
- **Phase 3:** Template mapping (sub-issues 2, 3) and variable names (sub-issue 4)
- **Phase 4:** Infrastructure (sub-issues 6, 7) — ASP.NET Core handler and repo linking

gautam-achieveai:
## 🏗️ Design Plan — Complete Source Mapping Support

[...]

PR #25, opened 2026-04-18
Phase 1: source-map core fixes and test foundation (#18)

### LINKED ISSUE #18: Add source map verification tooling and test suite
## Summary
There are currently **zero tests** for source map correctness in the NScript test suite. We need:
1. A tool/library that can parse `.map` files and validate their structure and mappings
2. A comprehensive test suite that compiles known C# inputs, parses the resulting `.map`, and asserts correct line/column mappings for every construct type

**Parent:** #12

## Acceptance Criteria
- [ ] Identify and integrate a .NET-compatible source map V3 parsing library, or write a minimal parser that can decode Base64 VLQ mappings back to structured `SourceMapping` objects
- [ ] Create a test project under `Test/Compiler/` (e.g., `SourceMap.Test`)
- [ ] Write test cases covering mappings for: method bodies, variable declarations, for/foreach/while loops, if/else branches, try/catch/finally, lambdas, async/await, property accessors
- [ ] Add test cases for XWML template bindings (once sub-issue for XWML mapping is addressed)
- [ ] Add test cases for Razor template expressions (once sub-issue for Razor mapping is addressed)
- [ ] Tests should verify forward mapping: given a C# source line, the corresponding JS line in the `.map` maps back to it
- [ ] Tests should verify structural validity of the `.map` file (valid JSON, correct version, sources array matches, VLQ segments decode correctly)
- [ ] Integrate into `dotnet test NScript_Full.sln`

## Notes
- This sub-issue should be tackled first (Phase 1) to establish the testing foundation before fixing mapping gaps
- Consider using an existing npm tool like `source-map` for cross-validation, or a .NET port
- The `SourceMap.ToString()` output includes a non-standard `sourcesLong` field — tests should verify both standard and NScript-specific fields

gautam-achieveai:
## Summary

Phase 1 of the source-mapping initiative tracked by #12. Establishes testable infrastructure and fixes known bugs in the core `SourceMap` class before Phases 2–4 build on it.

Closes #18.
Relates to #12.

### V3-spec compliance fixes (`Sources/Compiler/SourceMap/SourceMap.cs`)

- **Name index relative encoding.** `SourceMapping.ToStringRelative` previously emitted `SourceNameIndex` as an absolute VLQ. The V3 spec requires each emitted name index to be a delta from the **last emitted** name index. The fix tracks `previousNameIndex` as a separate accumulator in `SourceMap.ToString` so nameless segments do not poison the baseline. This unblocks Phase 3c (variable-name mapping).
- **Name array population.** `AddMapping` previously looked up the name in `this.names` but never added it, so the array was always empty. Now behaves like `files` — add on miss, reuse on hit.
- **Configurable `sourceRoot`.** `SourceMap.ToString` now honours the `SourceRoot` property, falling back to the legacy `{file}.ashx` default when not set. Keeps backward compatibility with existing deployments that use the bundled `SrcMapper.ashx` handler; unblocks Phases 4a/4b (modern handler + repo linking).

### Test seam (`Sources/Compiler/NScript.JST/JSWriter.cs`)

New public `WriteWithMap(TextWriter, string)` returns the generated `OwaSourceMapper.SourceMap` in-memory so tests can inspect mappings without parsing files from disk. The private `Write` overload now returns the `SourceMap`; the other public overloads ignore the return (no behavioural change for existing callers).

### Test project (`Test/Compiler/SourceMap.Test/`)

28 tests, all passing. Added to `NScript_Full.sln`.

- `Base64VLQTests` — known spec encodings + signed/unsigned round-trips.
- `SourceMapTests` — `sourceRoot` fallback vs override, names array population, and the name-index relative-encoding fix (the critical regression test).
- `JSWriterIntegrationTests` — end-to-end through the `WriteWithMap` seam, using a `DecodedMap` helper that reconstructs absolute mapping coordinates. Assertions are anchor-based (logical positions) rather than golden-file, so they survive token/formatting churn in later phases.
- `Base64VLQDecoder` — test-only VLQ decoder (production only needs the encoder).

## Key decisions

1. **Anchor-based integration tests, not golden `.map` files.** A golden-file approach would force every Phase 2/3 change to rewrite expected maps even when the mapping semantics are unchanged. The `DecodedMap` helper lets tests assert on `HasMappingFor(source, line)` directly.
2. **Track `previousNameIndex` separately from `previousMapping`.** The V3 spec defines each field's delta baseline as the last-emitted value of *that* field, not the prior segment's value. Mixing them (the pre-fix behaviour) breaks as soon as a segment without a name appears between two named segments.
3. **Keep `.ashx` as the fallback `sourceRoot`.** Any existing consumer that relied on the hard-coded behaviour continues to work unchanged when `SourceRoot` is left unset.

## Test plan

- [x] `dotnet test Test/Compiler/SourceMap.Test/SourceMap.Test.csproj -c Debug` — 28/28 passing
- [x] `dotnet build NScript_Full.sln -c Debug` — succeeds, no new warnings
- [x] `dotnet test Test/Compiler/NScript.Utils.Test/...` — 19/19 passing (unchanged)
- [x] `dotnet test Test/Compiler/NScript.JSParser.Test/...` — 41/41 passing (unchanged)
- [x] One unrelated pre-existing failure in `NScript.Converter.Test` (`TestAsyncAwait__TestExternalAwait`) confirmed to reproduce on the unchanged base branch — not introduced by this PR.


gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of Phase 1: source-map core fixes and test foundation (#18)

---

## Summary

The core V3 spec fixes (name-index delta encoding, names array population, sourceRoot fallback) are **correct and well-isolated**. The test foundation is solid — 28 tests pass, the Base64VLQ decoder is properly scoped to the test assembly, and the `WriteWithMap` test seam cleanly reuses the production code path.

**Two items need attention before merge.** The most important: the regression test for the primary fix (name-index delta encoding) is degenerate — it can't distinguish the old buggy code from the fix. See inline comments.

### Linked Work Items
- **#12**: Complete source mapping support for C#, XWML, and Razor templates
- **#18**: Add source map verification tooling and test suite

---

## Findings

### HIGH

| # | Finding | File |
|---|---------|------|
| 1 | Regression test for name-index delta is degenerate — passes with both absolute and delta encoding | `SourceMapTests.cs:106` |
| 2 | Solution file: SourceMap.Test nested under new top-level "Compiler" folder instead of existing `Test > Compiler` | `NScript_Full.sln` |

### MEDIUM

| # | Finding | File |
|---|---------|------|
| 3 | `sourceRoot` parameter in `Write(string, string)` is silently dropped — never wired to new `SourceMap.SourceRoot` | `JSWriter.cs:329` |
| 4 | Stale XML doc on private `Write` references `sourceRoot`/`mapFileName` params that don't exist | `JSWriter.cs:361-371` |
| 5 | `DecodedMap.Parse` uses hardcoded string markers — formatting changes cause silent false-passes | `JSWriterIntegrationTests.cs:149` |
| 6 | Missing edge case tests: duplicate name reuse, null File + null SourceRoot fallback | `SourceMapTests.cs` |

### LOW

| # | Finding | File |
|---|---------|------|
| 7 | O(n²) `List.IndexOf` for name dedup — will matter when Phase 3c wires up name mapping | `SourceMap.cs:228` |
| 8 | `WriteWithMap` doesn't accept `sourceRoot` — future Phase 4 tests can't exercise it end-to-end | `JSWriter.cs:356` |

---

## What's Well Done
- ✅ Clean test seam pattern — `WriteWithMap` reuses the exact production code path
- ✅ `Base64VLQDecoder` correctly scoped to test assembly only
- ✅ Layer boundaries respected (no cycles in project references)
- ✅ V3 name-index delta fix is architecturally sound — separate `previousNameIndex` tracker prevents nameless mapping poisoning
- ✅ SourceRoot fallback design is open/closed — existing behavior preserved while enabling new configurations
- ✅ Good test documentation with XML doc comments explaining intent and V3 spec rationale

---

**Review agents used:** architecture-review, exception-handling-review, test-coverage-review, performance-review, manual V3 spec analysis

Test/Compiler/SourceMap.Test/SourceMapTests.cs
```diff
@@ -0,0 +1,188 @@
...
+            string json = map.ToString();
+            string mappings = ExtractMappingsField(json);
+
+            var segments = DecodeFirstLineSegments(mappings);
+            Assert.AreEqual(2, segments.Count, "Expected two segments on the first line");
+
+            // Segment layout per V3 spec: [genCol, srcFile, srcLine, srcCol, nameIndex].
+            Assert.AreEqual(0, segments[0][4], "First mapping's name index is delta-from-zero = 0 ('first' at index 0).");
+            Assert.AreEqual(1, segments[1][4], "Second mapping's name index must be encoded as a DELTA (second - first = 1).");
+        }
```
gautam-achieveai, line 122: **[Gautam's review bot]** ⚠️ **HIGH — Degenerate regression test for the primary fix**

This test uses 2 names: `"first"` (index 0) and `"second"` (index 1). The delta values are `[0-0, 1-0] = [0, 1]` — but the **absolute** values (old buggy code) are also `[0, 1]`. The test passes with both the buggy and fixed implementations.

To truly validate delta encoding, you need 3+ distinct names where the third name's delta differs from its absolute index:

```csharp
[TestMethod]
public void ToString_ThreeDistinctNames_ThirdEncodedRelativeToSecond()
{
    var map = new SourceMap { File = "out.js" };
    map.AddMapping(0, 0, 0, 0, "Program.cs", name: "alpha");  // index 0
    map.AddMapping(0, 1, 0, 5, "Program.cs", name: "beta");   // index 1
    map.AddMapping(0, 2, 0, 10, "Program.cs", name: "gamma"); // index 2

    var segments = DecodeFirstLineSegments(ExtractMappingsField(map.ToString()));

    Assert.AreEqual(0, segments[0][4], "alpha: delta from 0 = 0");
    Assert.AreEqual(1, segments[1][4], "beta: delta from 0 = 1");
    // KEY: absolute would be 2; correct delta is 2-1 = 1
    Assert.AreEqual(1, segments[2][4], "gamma: delta from beta(1) = 1, NOT absolute 2");
}
```

The existing `ToString_NamelessMappingBetween_DoesNotPoisonNameIndexBaseline` has the same issue — with names at indices 0 and 1, both absolute and delta produce value 1 for the second name.

gautam-achieveai:
**[Gautam's review bot]** ⚠️ **HIGH — Solution folder nesting is incorrect** (`NScript_Full.sln`)

The new "Compiler" solution folder `{01A53387}` is created at the **top level** of the solution. But there's already a `Test > Compiler` folder at `{C3E4332C}` (under `Test` `{D77EDDB9}`) where other test projects like `RazorSkinParser.Test` and `NScript.Utils.Test` live.

SourceMap.Test ends up at the root level in Solution Explorer instead of under `Test > Compiler` alongside its siblings.

**Fix:** Remove the new "Compiler" solution folder entry entirely and change the nesting line to:
```
{A39BCDDB-B64F-4C2C-B081-E2279A6B5E21} = {C3E4332C-8AEF-4CE7-B1CF-030E7A42342C}
```

gautam-achieveai:
**[Gautam's review bot]** 📋 **Additional inline findings** (MEDIUM & LOW)

---

### MEDIUM — Stale doc comments + sourceRoot gap (`JSWriter.cs:361-371`)

The private `Write` method's XML comments reference `<param name="sourceRoot">` and `<param name="mapFileName">` but the actual parameters are `outputDirectory` and `outputMap`.

Additionally, the public `Write(string jsFileName, string sourceRoot)` at line 329 accepts `sourceRoot` from the production caller but never wires it to `sourceMapping.SourceRoot`. The new `SourceRoot` infrastructure in `SourceMap.cs` is dead on the production path.

If deferred to Phase 4: fix doc comments and add a `// TODO` so contributors know this is intentional.

---

### MEDIUM — Fragile JSON parsing in `DecodedMap.Parse` (`JSWriterIntegrationTests.cs:149-173`)

Uses hardcoded string markers (`"sourcesLong": ["`) and splits on `"\",\n\t\t\""`. If `SourceMap.ToString()` formatting changes even slightly, this parser silently returns empty results → tests false-pass.

If `IndexOf` returns -1, the subsequent `Substring` throws a cryptic `ArgumentOutOfRangeException`.

Suggestions:
- Add guard assertions: `Assert.IsTrue(srcEnd > srcStart, "sourcesLong not terminated")`
- After parsing, assert `result.Sources.Count > 0` when sources are expected

---

### MEDIUM — Missing edge case tests (`SourceMapTests.cs`)

1. **Duplicate name reuse**: No test verifies same name twice → single names array entry, correct index
2. **Null File + null SourceRoot**: The third fallback branch (`string.Empty`) is untested — `var map = new SourceMap()` with no File set

---

### LOW — O(n²) name lookup (`SourceMap.cs:228`)

`List<string>.IndexOf` is O(n) per call. When Phase 3c wires up name mapping, consider adding a parallel `Dictionary<string, int>` for O(1) lookups:

```csharp
private Dictionary<string, int> nameIndex = new();

if (!this.nameIndex.TryGetValue(name, out sNameIndex))
{
    sNameIndex = this.names.Count;
    this.names.Add(name);
    this.nameIndex[name] = sNameIndex;
}
```

gautam-achieveai (commented):
**[Gautam's review bot]** :white_check_mark: Incremental review of iteration 2 — **APPROVE**

---

All 7 findings from the initial review (2 HIGH + 5 MEDIUM) have been thoroughly addressed in `1588ec65`:

### HIGH — Fixed :white_check_mark:
| # | Finding | Resolution |
|---|---------|------------|
| 1 | Degenerate regression test (2 names can't distinguish delta from absolute) | Added `ThreeDistinctNames` test with 3 names where delta [0,1,1] ≠ absolute [0,1,2]. Also strengthened `NamelessMappingBetween` to use 3 named mappings. |
| 2 | Solution folder nesting (stray top-level "Compiler") | Removed `{01A53387}`, nesting now targets `{C3E4332C}` (existing Test > Compiler). Verified SourceMap.Test sits alongside RazorSkinParser.Test and NScript.Utils.Test. |

### MEDIUM — Fixed :white_check_mark:
| # | Finding | Resolution |
|---|---------|------------|
| 3 | `sourceRoot` silently dropped | Private `Write` now accepts `sourceRoot` param; public `Write(jsFileName, sourceRoot)` passes it through with `!string.IsNullOrEmpty` guard. |
| 4 | Stale XML doc comments | Completely rewritten to match the actual 5-parameter signature. |
| 5 | `DecodedMap.Parse` fragility | 3 guard assertions added with clear diagnostic messages for format changes. |
| 6 | Missing duplicate name reuse test | New `AddMapping_DuplicateName_ReusesSingleNamesEntry`. |
| 7 | Missing null File + null SourceRoot test | New `ToString_NoFileAndNoSourceRoot_FallsBackToEmptyString`. |

### Remaining LOW (acceptable, correctly deferred)
- O(n²) name lookup → Phase 3c
- WriteWithMap missing sourceRoot → Phase 4

### Build & Test
:white_check_mark: **31 tests pass** (28 original + 3 new)

No new issues found in the iteration 2 changes. Clean, focused commit that addresses every review item.

---

**Linked work items:** #12 (source mapping support), #18 (test tooling)

PR #28, opened 2026-04-19
Phase 3b (WI-14): Razor skin template source-map support

gautam-achieveai:
## Summary

Phase 3b of WI-12 (source-map support). Threads `IntermediateNode.Source` positions from Razor `.skin.cshtml` templates through the IR and into the emitted JST so the final V3 source map points back at the template file instead of reporting null positions.

- `TemplateIRBuilder.TryGetLocation` converts Razor's 0-based `LineIndex` to NScript's 1-based `Location`; the guard only filters truly synthetic negative indices (Razor `LineIndex == 0` survives as template line 1).
- Every IR node produced from a Razor intermediate node carries a `Location`: `SkinTemplateNode`, `HtmlNode`, `ExpressionBindingNode`, `ConditionalNode`, `LoopNode`, `EventNode`, `FunctionNode`, `SubControlNode`.
- `ExtractSubControlsFromHtml` takes an optional `hostLocation` threaded from the surrounding `HtmlContentIntermediateNode` so sub-controls pin to the line they appear on rather than collapsing to the `SkinTemplateNode` root.
- `GraphDescriptorJSTEmitter` gains an optional `fallbackLocation` ctor parameter (persisted in `_fallbackLocation`). ~20 `FunctionExpression` / `ReturnStatement` / `ExpressionStatement` construction sites now stamp the fallback location when no per-binding position is available. The nested `CollectionTopology` emitter prefers `ct.IrNode?.Location ?? _fallbackLocation` so loop-body graph descriptors retain per-loop-node lines.
- `RazorSkinJSTGenerator.GetTemplateLocation()` pulls the template-root `Location` and passes it into the emitter so skin-factory / getter expressions always have a file-scoped anchor.

Mirrors the XWML Phase 3a (WI-15) pattern for consistency.

## Test plan

- [x] `RazorSkinParser.Test` full suite: 171/171 pass.
- [x] `TemplateSourceMapTests` (new, 7 tests): IR root, HtmlNode, ExpressionBindingNode, ConditionalNode, LoopNode, EventNode all carry template-scoped `Location`; reflection-based contract check on `GraphDescriptorJSTEmitter`'s `fallbackLocation` ctor param + `_fallbackLocation` field.
- [x] `NScript_Full.sln` Release build: 0 errors, 0 new warnings.
- [x] Two rounds of `feature-dev:code-reviewer` — final verdict LGTM.

[...]

PR #27, opened 2026-04-19
Phase 3a (WI-15): XWML source-map support

gautam-achieveai:
[Gautam's Dev bot]

## Summary

Phase 3a of the source-map initiative (parent #12). Threads template `HtmlNode` positions through the XWML code-generation pipeline so the generated JavaScript carries source-map entries pointing back at the originating `.html` template rather than reporting null positions.

Closes #15. Parent tracking issue: #12.

## What changed

**`Sources/Compiler/XwmlParser/SkinCodeGenerator.cs`**
- Added `GetLocation(HtmlNode)` / `GetLocation(NodeInfo)` helpers that build a `Location` from the node's `Line` / `LinePosition`, returning `null` for synthetic nodes so those don't leak spurious map entries.
- `GenerateSkinFactory` / `GenerateSkinGetter` anchor their emitted `FunctionExpression`, `ExpressionStatement`, `MethodCallExpression`, `IfBlockStatement`, and `ReturnStatement` to the `<skin>` root's location.

**`Sources/Compiler/XwmlParser/NodeInfos/{HtmlNodeInfo,UIElementNodeInfo,TypeNodeInfo}.cs`**
- Threaded template `Location` onto the leftmost `IndexExpression` in generated assignments. `ExpressionStatement.CreateAssignmentExpression` picks up location from its left expression, so statement-level mapping flows automatically.
- `TypeNodeInfo`: new `GetCtorExpression(codeGenerator, location)` overload so ctors carry the template location; `GenerateStaticInitializers` threads location through property/field setters.

**`Sources/Compiler/XwmlParser/Binding/BinderInfo.cs`**
- `GenerateCode` now derives `Location` from the owning `NodeInfo` and sets it on the generated binder `MethodCallExpression` and its inner `IdentifierExpression`.

## Tests

New `Test/Compiler/XwmlParser.Test/XwmlSourceMapTests.cs` (3 tests, all passing):

1. **`Generate_TemplateStatementsCarryTemplateLocation`** — drives `TestArrayBinding.html` through the real `XwmlTemplatingPlugin` pipeline and asserts at least one JST node in the emitted statements carries a `Location` whose file is the template `.html`.
2. **`Generate_LocationsPointAtSkinOrChildLines`** — asserts at least one recorded location points at the `<skin>` or bound `<div>` lines (11-13) of `TestArrayBinding.html`. Guards against a future regression where locations could collapse to line 1.
[...]

Issue #19, opened 2026-04-10
Support linking source maps to ADO/GitHub repository files


gautam-achieveai:
## Summary
Instead of requiring local file access to serve source files, source maps should be able to reference files in Azure DevOps or GitHub repositories directly. This enables debugging production JS without access to the source code on disk — the browser fetches original sources from the repo's web interface.

**Parent:** #12

## Acceptance Criteria
- [ ] Investigate `sourceRoot` URL patterns that point to ADO/GitHub raw file endpoints (e.g., `https://raw.githubusercontent.com/{owner}/{repo}/{commit}/{path}` or ADO equivalent)
- [ ] Embed build metadata into the compiled DLL or alongside the `.map` file: repo origin URL, commit SHA, branch name
- [ ] The ASP.NET Core source handler (#17) should support proxying or redirecting to repo URLs when local files aren't available
- [ ] Support both Azure DevOps and GitHub repository URL formats
- [ ] Document how to configure repo-linked source maps

## Technical Notes
- Need a mechanism to capture `git remote get-url origin` and `git rev-parse HEAD` at build time and embed in the output
- Could use an assembly-level attribute, an MSBuild property, or a sidecar metadata file alongside the `.map`
- The `sourceRoot` field in the `.map` JSON controls the base URL for all source references
- For private repos, the source handler may need to proxy requests with authentication rather than redirect
- Consider how this interacts with the `sourcesLong` field that currently stores local absolute paths

## Open Questions
- Should the build embed repo metadata automatically, or require explicit configuration?
- How to handle authentication for private ADO/GitHub repos in the browser?
- Should we support both "direct browser fetch from repo" and "server-side proxy" modes?

gautam-achieveai:
[Gautam's Dev bot]

Starting analysis for WI-19 (repo-linked source maps). Will research sourceRoot handling, build metadata embedding options, and the ASP.NET Core source handler (from #16), then post a design plan.

gautam-achieveai:
[Gautam's Dev bot] **Implementation Plan for #19: Support linking source maps to ADO/GitHub repository files**

## Understanding

**Type:** Task / enhancement (parent: #12)
**Summary:** Make NScript-emitted `.map` files reference original sources in a remote repo (GitHub raw or Azure DevOps raw) so browsers can fetch sources without any local checkout. Opt-in at build time; backward-compatible with existing `.map` consumers.

**Acceptance Criteria (from issue):**
[...]

PR #30, opened 2026-04-21
WI-17: Populate source map names array with original C# identifiers

### LINKED ISSUE #17: Map original variable/symbol names in source maps
## Summary
The `SourceMap` class supports a `names` array and `SourceNameIndex` field in the V3 format, but no code ever populates them. After minification, JS variable names are unrecognizable — the source map should carry the original C# names so browser DevTools can display them.

**Parent:** #12

## Acceptance Criteria
- [ ] Investigate whether Source Map V3 name mappings restore original names in Chrome/Firefox DevTools (hover, watch, scope panel)
- [ ] If supported, populate the `names` array in `SourceMap` with original C# identifiers during `JSWriter` emission
- [ ] Pass original identifier names from JST `IdentifierToken` nodes to `AddMapping()` calls
- [ ] Verify that DevTools displays original names when hovering over minified variables
- [ ] If name mapping has limited browser support, document the limitations and what does work

## Technical Notes
- `SourceMap.AddMapping()` already accepts an optional `name` parameter (line 202) but it's never called with one
- `SourceMapping.SourceNameIndex` and `SourceMapping.ToStringRelative()` already handle name encoding
- The JST `IdentifierToken` likely needs to carry both the minified and original name
- Source Map V3 spec: name mappings appear as the 5th field in a VLQ segment

gautam-achieveai:
[Gautam's Dev bot]

Closes #17. Phase 3c of WI-12 (parent).

## Summary
The `SourceMap` class already supported the V3 `names` array and `SourceNameIndex` encoding — what was missing was ever **populating** it. After minification, JS variable names are unrecognizable unless the source map tells DevTools what the original names were. This PR threads `IIdentifier.OriginalSuggestedName` through `JSWriter` into `SourceMap.AddMapping(..., name:)` so the array gets populated.

## Changes
1. **`GenericStrToken.cs`** — added an optional `OriginalName` property carrying the pre-minification identifier name alongside the emitted text. New 4-arg constructor; existing 3-arg constructor preserved for all other callers.
2. **`JSWriter.Write(IIdentifier)`** — populates the token's `OriginalName` from the identifier's `OriginalSuggestedName` whenever the scope has renamed it (i.e. `emitted != original`). Empty/whitespace originals are skipped to avoid bogus blank entries in the array.
3. **`JSWriter` emission loop** — when adding a mapping for an identifier token, forwards the token's `OriginalName` to `SourceMap.AddMapping(..., name:)` so the index is recorded correctly in the V3 segment.

## Filter rule — only emit names that actually help
A name is recorded only when:
- The token is an identifier (`TokenType.IdentifierToken`).
- `OriginalSuggestedName` is non-null and non-whitespace.
- `OriginalSuggestedName != GetName()` — i.e. the scope renamed it.

This keeps the `names` array proportional to actual rename events. Non-minified builds where the emitted text already equals the source name skip the array entirely.

## Tests
New `Test/Compiler/SourceMap.Test/VariableNameMappingTests.cs` (5 cases), all green alongside the existing 46 suite tests:

- Renamed identifier populates `names` with the original.
- Unrenamed identifier does not populate (map stays minimal).
- Two renamed identifiers sharing an original collapse to one `names` entry.
- Raw `WriteIdentifier(string)` has no `IIdentifier` context → no entry.
- Keyword tokens never write to `names` (type-dispatch guard).

`SourceMap.Test`: **51/51 pass**. Full solution builds clean (0 errors).

## Acceptance criteria mapping (per #17)
- [x] Investigate browser support — V3 name mappings are used by Chrome/Firefox DevTools for scope/hover display (spec-standard behavior).
- [x] Populate `names` array in `SourceMap` during `JSWriter` emission.
- [x] Pass original names from identifier tokens to `AddMapping()`.
- [ ] End-user DevTools verification — requires a live app; automated suite verifies the emitted JSON structure, which is what DevTools consumes.
- [x] Limitations documented: names are only emitted when rename actually changes the symbol; raw-string `WriteIdentifier` callers don't contribute names.

## Related
Parent issue: #12 (tracking source-map completeness)
Previous phases: #18 (PR #25), #13 (PR #26), #15 (PR #27), #14 (PR #28)
Next phases after this: #16 (SrcMapper.ashx replacement), #19 (repo linking)

### DIFF
--- a/Test/Compiler/SourceMap.Test/VariableNameMappingTests.cs
+++ b/Test/Compiler/SourceMap.Test/VariableNameMappingTests.cs
@@ -160,5 +160,137 @@ namespace OwaSourceMapper.Test
                 map.ToString().Contains("\"names\":"),
                 "Keyword tokens must not add entries to the names array.");
         }
+
+        /// <summary>
+        /// Strong structural check — decodes the VLQ <c>mappings</c> payload and confirms
+        /// the renamed-identifier segment has exactly 5 fields and that the 5th field
+        /// (name index, delta-encoded) is 0 for the first-and-only name. Without this
+        /// test the suite asserts only on the <c>names</c> array contents and would miss
+        /// a regression where the name index was dropped from the VLQ segment itself.
+        /// </summary>
+        [TestMethod]
+        public void WriteWithMap_RenamedIdentifier_MappingSegmentEncodesNameIndex()
+        {
+            var scope = new IdentifierScope(isExecutionScope: true);
+            SimpleIdentifier.CreateScopeIdentifier(scope, "counter", enforceSuggestion: false);
+            var renamed = SimpleIdentifier.CreateScopeIdentifier(scope, "counter", enforceSuggestion: false);
+
+            var writer = new JSWriter(isIndented: false, isOptimized: false);
+            writer.EnterLocation(new Location("Program.cs", 10, 5, 10, 13));
+            writer.Write(renamed);
+            writer.LeaveLocation();
+
+            using var stringWriter = new StringWriter();
+            var map = writer.WriteWithMap(stringWriter, "out.js");
+            string json = map.ToString();
+
+            var segments = DecodeFirstLineSegments(ExtractMappingsField(json));
+            Assert.AreEqual(1, segments.Count, "Expected one mapping segment on the first line.\n" + json);
+            Assert.AreEqual(
+                5,
+                segments[0].Count,
+                "Renamed identifier's segment must carry all 5 V3 fields (genCol, srcFile, srcLine, srcCol, nameIndex).\n" + json);
+            Assert.AreEqual(
+                0,
+                segments[0][4],
+                "First-and-only name: delta from baseline 0 = 0.\n" + json);
+        }
+
+        /// <summary>
+        /// Integration-level regression for the Phase 1 <c>previousNameIndex</c> fix —
+        /// drives it through <see cref="JSWriter.Write(IIdentifier)"/> rather than direct
+        /// <see cref="SourceMap.AddMapping"/> calls. Writes two renamed identifiers with
+        /// a keyword between them; the second identifier's name-index delta must be
+        /// encoded relative to the first identifier's name index, not poisoned by the
+        /// intervening nameless keyword mapping.
+        /// </summary>
+        [TestMethod]
+        public void WriteWithMap_KeywordBetweenRenamedIdentifiers_NameIndexBaselinePreserved()
+        {
+            var scope = new IdentifierScope(isExecutionScope: true);
+
+            // Two separate collision pairs → two distinct renamed identifiers with
+            // different original names ("alpha" and "beta").
+            SimpleIdentifier.CreateScopeIdentifier(scope, "alpha", enforceSuggestion: false);
+            var renamedAlpha = SimpleIdentifier.CreateScopeIdentifier(scope, "alpha", enforceSuggestion: false);
+            SimpleIdentifier.CreateScopeIdentifier(scope, "beta", enforceSuggestion: false);
+            var renamedBeta = SimpleIdentifier.CreateScopeIdentifier(scope, "beta", enforceSuggestion: false);
+
+            Assert.AreNotEqual(renamedAlpha.OriginalSuggestedName, renamedAlpha.GetName());
+            Assert.AreNotEqual(renamedBeta.OriginalSuggestedName, renamedBeta.GetName());
+
+            var writer = new JSWriter(isIndented: false, isOptimized: false);
+            writer.EnterLocation(new Location("Program.cs", 1, 0, 1, 5));
+            writer.Write(renamedAlpha);
+            writer.LeaveLocation();
+            writer.EnterLocation(new Location("Program.cs", 2, 0, 2, 6));
+            writer.Write(Keyword.Return);
+            writer.LeaveLocation();
+            writer.EnterLocation(new Location("Program.cs", 3, 0, 3, 4));
+            writer.Write(renamedBeta);
+            writer.LeaveLocation();
+
+            using var stringWriter = new StringWriter();
+            var map = writer.WriteWithMap(stringWriter, "out.js");
+            string json = map.ToString();
+
+            Assert.IsTrue(
+                json.Contains("\"names\": [\"alpha\",\"beta\"]")
+                    || json.Contains("\"names\": [\"alpha\", \"beta\"]"),
+                "Expected names array containing alpha and beta in the JSON output.");
+
+            var segments = DecodeFirstLineSegments(ExtractMappingsField(json));
+            Assert.AreEqual(3, segments.Count, "Expected three segments: alpha, keyword, beta.");
+            Assert.AreEqual(5, segments[0].Count, "Renamed alpha segment must carry a name field.");
+            Assert.AreEqual(4, segments[1].Count, "Keyword segment must NOT carry a name field.");
+            Assert.AreEqual(5, segments[2].Count, "Renamed beta segment must carry a name field.");
+            Assert.AreEqual(0, segments[0][4], "alpha: delta from baseline 0 = 0.");
+            Assert.AreEqual(
+                1,
+                segments[2][4],
+                "beta: delta from last EMITTED name index (alpha=0) = 1. "
+                + "Intervening nameless keyword mapping must not poison the baseline.");
+        }
+
+        private static string ExtractMappingsField(string json)
+        {
+            const string marker = "\"mappings\": \"";
+            int start = json.IndexOf(marker, System.StringComparison.Ordinal);
+            Assert.IsTrue(start >= 0, "mappings field not found in:\n" + json);
+            start += marker.Length;
+            int end = json.IndexOf('"', start);
+            Assert.IsTrue(end > start, "mappings field not terminated in:\n" + json);
+            return json.Substring(start, end - start);
+        }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of PR #30 — WI-17: Populate source map names array with original C# identifiers (Phase 3c of #12 source mapping effort).

## Summary

Clean, well-scoped change. The data flow is correct: `IIdentifier.OriginalSuggestedName` → `GenericStrToken.OriginalName` → token loop extraction → `SourceMap.AddMapping(name:)`. The design choice of carrying the original name on the token is consistent with the existing `Location` precedent on `TokenBase`. Backward compatibility is maintained via constructor chaining.

**Agents used:** architecture-review, performance-review, test-coverage-review

## Findings

### 🟡 [MEDIUM] Tests verify `names` array but not that VLQ mappings reference it
All 5 tests assert on the `"names"` JSON array contents (e.g., `"names": ["counter"]"`), but none verify that any mapping **segment** in the VLQ `"mappings"` string contains a 5th field referencing a name index. A bug where `SourceMapping` stores `NameIndex = -1` despite receiving a valid name would leave the names array present-but-orphaned — DevTools would see the list of names but never associate any symbol with one.

**Suggestion:** The existing `DecodedMap` helper in `JSWriterIntegrationTests.cs` already decodes `NameIndex` from VLQ segments. For the happy-path renamed-identifier test, parse the map and assert at least one mapping has `NameIndex >= 0`.

### 🔵 [LOW] No test for CompoundIdentifier with renamed sub-identifier
All tests use `SimpleIdentifier` directly. `CompoundIdentifier` (member access like `obj.field`) takes the `else` branch of `Write(IIdentifier)` and exercises the recursive path. Since recursion delegates to the `SimpleIdentifier` branch which is tested, this is low risk — but it's the most common expression shape in a C#-to-JS transpiler and worth a quick integration test to guard against future refactoring.

### 🔵 [LOW] No test for two different original names (delta encoding)
Test 3 verifies dedup (same name twice → one entry). No test writes two identifiers with *different* original names (e.g., `"counter"` and `"total"`) to exercise the VLQ delta-encoding of a second distinct name index. Delta encoding bugs are subtle — off-by-one in baseline tracking would cause every mapping after the first to reference the wrong name.

### ⚪ [NIT] Unused using directive
`using System.Collections.Generic;` in `VariableNameMappingTests.cs` appears unused.

## ✅ What's Good
- Surgical 3-file change, well-documented commit message
- `GetOriginalNameForMap` correctly skips null/whitespace and same-as-emitted names (avoids map bloat)
- Precondition asserts in tests prevent silent false-passes if scope renaming behavior changes
- No new project dependencies, consistent layering
- No performance concerns (build-time tool, names array kept small by design)

## Verdict: ✅ APPROVE

Test/Compiler/SourceMap.Test/VariableNameMappingTests.cs
```diff
@@ -0,0 +1,164 @@
...
+            writer.LeaveLocation();
+
+            using var stringWriter = new StringWriter();
+            var map = writer.WriteWithMap(stringWriter, "out.js");
+            string json = map.ToString();
+
+            StringAssert.Contains(
+                json,
+                "\"names\": [\"counter\"]",
+                "Pre-minification identifier name must appear in the names array.\n" + json);
```
gautam-achieveai, line 53: **[Gautam's review bot]** 🟡 **[MEDIUM]** The assertion here verifies the `names` array JSON but doesn't verify that any VLQ mapping **segment** actually references a name index (5th VLQ field). A bug where `SourceMapping` stores `NameIndex = -1` despite receiving a valid name would leave the names array present-but-orphaned.

Suggestion: reuse the `DecodedMap` helper from `JSWriterIntegrationTests.cs` to also assert:
```csharp
var decoded = DecodedMap.Parse(json);
Assert.IsTrue(
    decoded.Mappings.Exists(m => m.NameIndex >= 0),
    "At least one mapping segment must reference a name index.");
```

gautam-achieveai (commented):
**[Gautam's review bot]** Incremental re-review of iteration 2 (commit 9d2bd09).

## Summary

Iteration 2 adds two new tests and two VLQ-decoding helpers that directly address the MEDIUM finding from iteration 1. All 7 tests pass. No new issues found.

## Previous Findings — Status

| 1 | MEDIUM | VLQ segments don't verify name index reference | **Addressed** — `WriteWithMap_RenamedIdentifier_MappingSegmentEncodesNameIndex` decodes VLQ, asserts 5 fields, verifies name-index delta = 0 |
| 2 | LOW | No delta encoding test with different names | **Addressed** — `WriteWithMap_KeywordBetweenRenamedIdentifiers_NameIndexBaselinePreserved` verifies delta = 1 for second name, confirms keyword doesn't poison baseline |
| 3 | LOW | No CompoundIdentifier test | Deferred (acceptable — recursive delegation to tested SimpleIdentifier path) |
| 4 | NIT | Unused `using System.Collections.Generic` | **Resolved** — now used by `List<List<int>>` in new helpers |

## Notes on Implementation Choice

The developer implemented their own `ExtractMappingsField` + `DecodeFirstLineSegments` helpers using `Base64VLQDecoder` rather than reusing the `DecodedMap` helper from `JSWriterIntegrationTests.cs`. This is a reasonable approach — it provides more granular control (asserting exact field counts and specific delta values per segment), and keeps the test file self-contained.

## Verdict: APPROVE

All substantive feedback from iteration 1 has been addressed. The VLQ-level assertions are actually stronger than what was suggested — they verify exact segment field counts and specific delta values, not just at least one mapping has a name index. Well done.

Test/Compiler/SourceMap.Test/VariableNameMappingTests.cs
gautam-achieveai: **[Gautam's review bot]** This is addressed in iteration 2. The new `WriteWithMap_RenamedIdentifier_MappingSegmentEncodesNameIndex` test decodes the VLQ mappings payload and asserts: (1) segment has exactly 5 V3 fields, and (2) the 5th field (name index delta) equals 0 for the first-and-only name.

The implementation uses a self-contained `DecodeFirstLineSegments` helper with `Base64VLQDecoder` rather than `DecodedMap` — this gives even more precise assertions (exact field count + specific delta values per segment).

Additionally, `WriteWithMap_KeywordBetweenRenamedIdentifiers_NameIndexBaselinePreserved` covers the delta-encoding path with interleaved nameless mappings, verifying the baseline isn't poisoned.

gautam-achieveai:
Please address CompoundIdentifier Tests. These are interesting cases.

gautam-achieveai:
[Gautam's Dev bot]

Addressed in commit 2c173537 — added two CompoundIdentifier tests to `VariableNameMappingTests.cs`:

1. **`WriteWithMap_CompoundIdentifier_RenamedLeadingPart_PopulatesNamesArray`** — exercises the `else` branch of `JSWriter.Write(IIdentifier)` with a renamed leading sub-identifier (`counter.field`). Verifies (a) the original name flows into `"names"` via the recursive `SimpleIdentifier` path and (b) the VLQ segment carries the 5th (name-index) field at delta 0.

2. **`WriteWithMap_CompoundIdentifierParts_PerPartLocation_BothNamesRecorded`** — covers the real-codegen pattern where each member-access part is written under its own `EnterLocation` span. Verifies both renamed sub-identifiers contribute entries to the names array.

Interesting case surfaced while writing the tests: `Write(CompoundIdentifier)` uses a single top-location for all sub-tokens, so when the whole compound is wrapped by one `EnterLocation`, only the first sub-identifier's mapping is emitted (subsequent tokens share the location and are coalesced). Test 2 documents the per-part-span pattern that callers should use when each sub-identifier has distinct source positions.

All 9 tests pass locally.

gautam-achieveai (commented):
**[Gautam's review bot]** ✅ **Iteration 3 — APPROVE**

**Incremental review** of commit `2c17353` (+82 lines, test-only).

### What changed
Two new CompoundIdentifier tests addressing the deferred LOW finding from iteration 1 and the explicit reviewer request in the thread:

1. **`WriteWithMap_CompoundIdentifier_RenamedLeadingPart_PopulatesNamesArray`** — Exercises the `else` branch of `Write(IIdentifier)` with `CompoundIdentifier(renamedCounter, field)`. Includes VLQ segment assertions (5-field check + delta=0). Solid coverage of the recursive delegation path.

2. **`WriteWithMap_CompoundIdentifierParts_PerPartLocation_BothNamesRecorded`** — Documents the real-codegen pattern where each member-access part is written under its own `EnterLocation` span, verifying both renamed sub-identifiers contribute to the names array.

### Findings status
| 1 | MEDIUM | VLQ name index verification | ✅ Addressed (iter 2) |
| 2 | LOW | CompoundIdentifier coverage | ✅ **Addressed (iter 3)** |
| 3 | LOW | Delta encoding coverage | ✅ Addressed (iter 2) |
| 4 | NIT | Unused using | ✅ Resolved (iter 2) |

**All 4 original findings are now addressed. All 9 tests pass. No new issues found.**

### Work item context
- **User Story #17**: Map original variable/symbol names in source maps
- Phase 3c of #12 source mapping effort

Clean, well-tested implementation. Ship it. 🚢

gautam-achieveai:
**[Gautam's review bot]** Re: CompoundIdentifier tests request — addressed in commit 2c17353. Both new tests exercise the `CompoundIdentifier` path through `Write(IIdentifier)` and verify original names flow correctly into the source map. The `PerPartLocation` variant documents an important behavioral nuance about location coalescing. All 9 tests pass. ✅

PR #39, opened 2026-04-24
[Gautam's Dev bot] WI-37: E2E tests for SourceMap.Server handler

### LINKED ISSUE #37: E2E test for new ASP.NET Core SourceMap.Server handler (WebApplicationFactory)
**Parent umbrella:** #12 (Complete source mapping support)
**Dependency:** Follow-up on #16 (Replace SrcMapper.ashx with ASP.NET Core handler — PR #36)

## Context

PR #36 ships the new `SourceMap.Server` handler and end-to-end compiler plumbing (`-sourceMapRoot`, `<SourceMapRoot>`, `SourceMap.EmitLegacyAshxHandler`), but its test suite is deliberately scoped to:

- `SourceMap.Server.Test` (24 tests) — drives the handler via `TestServer`/`HttpContext` (unit-level on the handler itself).
- `SourceMap.Test` (60 tests) — asserts `Write()` honors `EmitLegacyAshxHandler` and the `JSWriter` toggle logic.

Neither stands up a full host, so we don't have automated coverage for the **browser-DevTools-resolves-C#-through-the-new-handler** scenario end-to-end.

## Scope

Add an end-to-end test (leveraging `Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory<T>`) that:

1. Boots an in-process ASP.NET Core server that calls `endpoints.MapSourceMapFiles(...)` against a `MapsDirectory` populated with real compiler output from a small NScript fixture project.
2. Compiles the fixture with the new `-sourceMapRoot` flag so the generated `.js.map` points to the new endpoint (not `SrcMapper.ashx`).
3. Issues HTTP GETs against `{prefix}/{mapName}/{shortName}` for the mapped short names that appear in the map's `sources` array and asserts:
   - `200 OK`, `text/plain; charset=utf-8`, body equals the original C#/XWML/Razor source bytes.
   - `404` for a short name not in the map.
   - `400` for traversal attempts in `sourceName` and for invalid `mapName` patterns.
   - `AllowedSourceRoots` containment works against a real on-disk layout (sibling-prefix regression — e.g., `<work>/src` allowed, `<work>/src-evil` rejected).
4. Verifies the `run-qunit.mjs` dev harness route (`/sourcemap/{mapName}/{shortName}`) resolves correctly against the same fixture (keeps local `qunit` source map debugging working).

## Deliverables

- New test project (or extended `SourceMap.Server.Test`) using `WebApplicationFactory<TEntryPoint>` with a minimal `Program.cs`.
- Fixture NScript project checked in under `Test/Framework/` (tiny — a couple of C# files, one XWML template, one Razor skin) or generated on-the-fly via existing test harness helpers if that pattern is already used in this repo.
- CI green on the new test project.

## Acceptance criteria

- [ ] E2E test proves a compiler-emitted `.js.map` whose `sourceRoot` points at the new handler can be resolved by an HTTP client (simulating Chrome DevTools) against a running `SourceMap.Server` endpoint.
- [ ] Negative paths (404/400, containment violation) exercised against the same live host.
- [ ] Test runs as part of the solution's test phase — no manual setup required.
- [ ] PR #36 can be merged once this issue is implemented (per maintainer request on PR #36 2026-04-21T20:48:04Z).

## Notes

- Option discussed on PR #36 was `WebApplicationFactory`-style e2e versus manual cURL; this issue formalizes the former.
- Do **not** re-open the handler's unit-level surface from this test — keep it a black-box HTTP integration test.

gautam-achieveai:
[Gautam's Dev bot]

Fixes #37. Follows up on #36 (the handler itself) with three complementary end-to-end tiers so the public contract is verified against routing, URL decoding, streaming, and the actual browser resolver — not just the handler unit test bed.

## What's covered

1. **`WebApplicationFactory<TestStartup>` suite** (`Test/Compiler/SourceMap.Server.Test/EndToEndServerTests.cs`, 10 new test methods):
   - Byte-exact round-trips for `.cs`, `.xwml`, `.cshtml` fixtures
   - Unknown short name / unknown map → 404
   - `mapName` whitelist rejection → 400 (`a:b`, `a b`, `.hidden`)
   - Dotted-traversal path normalization — accepts any non-200 since ASP.NET Core collapses `..` before routing, and asserts no fixture content ever leaks
   - Sibling-prefix containment regression (`<work>/src` allow-list vs. `<work>/src-evil/Secret.cs` target) at the full HTTP transport layer
   - `sourceRoot` wired into the emitted map + no legacy `.ashx` sidecar
   - Canonical AC #1 DevTools flow: compose `{sourceRoot}/{shortName}` and assert the real pipeline streams back the fixture bytes

2. **`SourceMap.Server.TestHost` console app** (new project) — emits a compiler-shaped fixture via the real `OwaSourceMapper.SourceMap` and boots Kestrel on an ephemeral port so a Playwright browser test can drive the same handler end-to-end. Decoupled from the MSTest project so the Exe doesn't pull in test frameworks.

3. **Node + Playwright tests** under `Test/Framework/TestWebApplication`:
   - `source-map-route.mjs` — extracted from `run-qunit.mjs` so the dev-ergonomic helper is reusable and unit-testable
   - `source-map-route.test.mjs` — 6 `node:test` smoke cases covering the same 200/404/400 shapes
   - `source-map-browser.test.mjs` — Playwright + Chromium test that loads `/fixture.html` (served by the TestHost), fetches `{sourceRoot}/{shortName}` from the page context, and asserts byte-identical fixture bytes come back through the real C# handler. Skips cleanly when the TestHost binary or Playwright isn't available.

## Forward compatibility for WI-19

Per the maintainer ask on the plan, `FixtureMapEmitter.Emit(..., sourceRoot)` accepts any URL — including a future `https://raw.githubusercontent.com/…` — and threads it through unchanged. The handler's `sources[]` resolution is independent of `sourceRoot`, so WI-19 can swap in a repo-link frontend without touching either the emitter or the handler.

## Test results

- `SourceMap.Server.Test`: **44/44 passed** (34 existing + 10 new)
- `SourceMap.Test`: **60/60 passed**
- `source-map-route.test.mjs`: **6/6 passed**
- `source-map-browser.test.mjs`: **1/1 passed**
- Full `dotnet build NScript_Full.sln -c Release`: clean (0 errors, 189 pre-existing warnings)

Part of the parent source-mapping initiative #12.

### DIFF
--- a/Test/Compiler/SourceMap.Server.Test/TestStartup.cs
+++ b/Test/Compiler/SourceMap.Server.Test/TestStartup.cs
@@ -1,0 +1,100 @@
+// -----------------------------------------------------------------------
+// <copyright file="TestStartup.cs" company="">
+//     Copyright (c) . All rights reserved.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace OwaSourceMapper.Server.Test
+{
+    using System;
+    using System.Threading;
+    using Microsoft.AspNetCore.Builder;
+    using Microsoft.AspNetCore.Hosting;
+    using Microsoft.AspNetCore.Mvc.Testing;
+    using Microsoft.Extensions.Hosting;
+
+    /// <summary>
+    /// Program/Startup hybrid used as <c>TEntryPoint</c> for
+    /// <see cref="Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory{TEntryPoint}"/>
+    /// in the SourceMap.Server end-to-end test suite. The type must be
+    /// <c>public</c> so the factory can locate it via reflection.
+    /// </summary>
+    /// <remarks>
+    /// Each test swaps its own <see cref="SourceMapFileHandlerOptions"/> onto
+    /// <see cref="CurrentOptions"/> in <c>TestInitialize</c>, then constructs a
+    /// fresh <see cref="Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory{TEntryPoint}"/>.
+    /// <c>Main</c> pulls the latched options at request time, so each factory
+    /// instance sees the options its test configured.
+    /// </remarks>
+    public class TestStartup
+    {
+        private static SourceMapFileHandlerOptions currentOptions;
+
+        /// <summary>
+        /// Options used by the endpoint on the next pipeline build. Swapped
+        /// atomically so race conditions between parallel test runs cannot
+        /// leak configuration across factories.
+        /// </summary>
+        public static SourceMapFileHandlerOptions CurrentOptions
+        {
+            get => Volatile.Read(ref currentOptions);
+            set => Volatile.Write(ref currentOptions, value);
+        }
+
+        /// <summary>
+        /// Route prefix used when registering the handler. Must match what the
+        /// tests issue HTTP requests against.
+        /// </summary>
+        public const string PathPrefix = "/sourcemap";
+
+        /// <summary>
+        /// Entry point invoked by <see cref="Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory{TEntryPoint}"/>.
+        /// Kept minimal: routing + the one endpoint under test, nothing else.
+        /// </summary>
+        /// <param name="args"> Process args (unused). </param>
+        public static void Main(string[] args)
+        {
+            // WebApplicationFactory probes the entry assembly for its content
+            // root; set it explicitly so `WebApplication.CreateBuilder` does
+            // not try to locate wwwroot relative to a non-existent directory
+            // built from the assembly's simple name.
+            var builder = WebApplication.CreateBuilder(new WebApplicationOptions
+            {
+                Args = args,
+                ContentRootPath = AppContext.BaseDirectory,
+            });
+            var app = builder.Build();
+            var options = CurrentOptions ?? new SourceMapFileHandlerOptions();
+            app.MapSourceMapFiles(PathPrefix, options);
+            app.Run();
+        }
+    }
+
+    /// <summary>
+    /// Thin wrapper around <see cref="WebApplicationFactory{TEntryPoint}"/>
+    /// that pins the content root to <see cref="AppContext.BaseDirectory"/>.
+    /// </summary>
+    /// <remarks>
+    /// The default content-root discovery logic of the factory prefers the
+    /// solution-relative path derived from the entry-point type's simple
+    /// assembly name, which does not exist in our layout. Hosting the app in
+    /// the test bin directory keeps fixtures resolvable via
+    /// <c>AppContext.BaseDirectory</c> and matches the path used by the
+    /// production <c>SourceMapFileHandler</c> when deployed.
+    /// </remarks>
+    public sealed class SourceMapServerFactory : WebApplicationFactory<TestStartup>
+    {
+        protected override IHostBuilder CreateHostBuilder()
+        {
+            // Returning null forces the factory to build the host via the
+            // Main-invocation path (HostFactoryResolver), where our
+            // WebApplicationOptions.ContentRootPath takes effect.
+            return null;
+        }
+
+        protected override void ConfigureWebHost(IWebHostBuilder builder)
+        {
+            builder.UseContentRoot(AppContext.BaseDirectory);
+        }
+    }
+}
--- a/Test/Compiler/SourceMap.Server.TestHost/Program.cs
+++ b/Test/Compiler/SourceMap.Server.TestHost/Program.cs
@@ -1,0 +1,221 @@
+// -----------------------------------------------------------------------
+// <copyright file="Program.cs" company="">
+//     Copyright (c) . All rights reserved.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace OwaSourceMapper.Server.TestHost
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using Microsoft.AspNetCore.Builder;
+    using Microsoft.AspNetCore.Hosting;
+    using Microsoft.AspNetCore.Hosting.Server;
+    using Microsoft.AspNetCore.Hosting.Server.Features;
+    using Microsoft.Extensions.DependencyInjection;
+    using OwaSourceMapper.Server;
+
+    /// <summary>
+    /// End-to-end test host for the Playwright browser suite. Two subcommands:
+    /// <list type="bullet">
+    ///   <item>
+    ///     <term><c>emit &lt;workDir&gt;</c></term>
+    ///     <description>Produces a compiler-shaped <c>{mapName}.map</c> + companion
+    ///       <c>{mapName}.js</c> under <c>{workDir}/maps</c>, and mirrors the
+    ///       bundled <c>Fixtures</c> directory into <c>{workDir}/src</c> so the
+    ///       map's <c>sourcesLong</c> entries resolve to real bytes.</description>
+    ///   </item>
+    ///   <item>
+    ///     <term><c>serve &lt;workDir&gt;</c></term>
+    ///     <description>Starts Kestrel on an ephemeral port, wires
+    ///       <c>/sourcemap</c> to <see cref="SourceMapFileHandler"/>, serves the
+    ///       <c>{workDir}/maps</c> directory as static content (for the
+    ///       <c>.js</c> + <c>.map</c> pair), and prints <c>LISTENING
+    ///       http://127.0.0.1:{port}</c> on stdout so the node harness can parse
+    ///       the URL and drive Playwright against it.</description>
+    ///   </item>
+    /// </list>
+    /// The two subcommands are deliberately decoupled so the node test can
+    /// emit fixtures up front (cheap, deterministic) and only pay the Kestrel
+    /// startup cost once per scenario group.
+    /// </summary>
+    public static class Program
+    {
+        private const string DefaultMapName = "app";
+        private const string PathPrefix = "/sourcemap";
+
+        public static int Main(string[] args)
+        {
+            if (args == null || args.Length == 0)
+            {
+                Console.Error.WriteLine("Usage: SourceMap.Server.TestHost (emit|serve) <workDir>");
+                return 2;
+            }
+
+            string cmd = args[0];
+            try
+            {
+                return cmd switch
+                {
+                    "emit" => RunEmit(args),
+                    "serve" => RunServe(args),
+                    _ => InvalidUsage(cmd),
+                };
+            }
+            catch (Exception ex)
+            {
+                Console.Error.WriteLine("TestHost error: " + ex.Message);
+                Console.Error.WriteLine(ex.StackTrace);
+                return 1;
+            }
+        }
+
+        private static int InvalidUsage(string cmd)
+        {
+            Console.Error.WriteLine("Unknown subcommand: " + cmd);
+            Console.Error.WriteLine("Expected: emit | serve");
+            return 2;
+        }
+
+        private static int RunEmit(string[] args)
+        {
+            if (args.Length < 2)
+            {
+                Console.Error.WriteLine("emit requires <workDir>");
+                return 2;
+            }
+
+            string workDir = Path.GetFullPath(args[1]);
+            string mapsDir = Path.Combine(workDir, "maps");
+            string srcDir = Path.Combine(workDir, "src");
+            Directory.CreateDirectory(mapsDir);
+            Directory.CreateDirectory(srcDir);
+
+            string fixtureRoot = Path.Combine(AppContext.BaseDirectory, "Fixtures");
+            var fixtures = new List<string>();
+            foreach (var rel in new[] { "Program.cs", "View.xwml", "Skin.cshtml" })
+            {
+                string src = Path.Combine(fixtureRoot, rel);
+                if (!File.Exists(src))
+                {
+                    Console.Error.WriteLine("Missing fixture: " + src);
+                    return 3;
+                }
+
+                string dst = Path.Combine(srcDir, rel);
+                File.Copy(src, dst, overwrite: true);
+                fixtures.Add(dst);
+            }
+
+            EmitMap(DefaultMapName, mapsDir, fixtures, sourceRoot: PathPrefix + "/" + DefaultMapName);
+            Console.WriteLine("EMITTED " + Path.Combine(mapsDir, DefaultMapName + ".map"));
+            return 0;
+        }
+
+        private static int RunServe(string[] args)
+        {
+            if (args.Length < 2)
+            {
+                Console.Error.WriteLine("serve requires <workDir>");
+                return 2;
+            }
+
+            string workDir = Path.GetFullPath(args[1]);
+            string mapsDir = Path.Combine(workDir, "maps");
+            string srcDir = Path.Combine(workDir, "src");
+
+            if (!Directory.Exists(mapsDir))
+            {
+                Console.Error.WriteLine("maps dir missing; run `emit` first: " + mapsDir);
+                return 3;
+            }
+
+            var builder = WebApplication.CreateBuilder(new WebApplicationOptions
+            {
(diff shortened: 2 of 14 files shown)


gautam-achieveai:
[Gautam's Dev bot]

Self-review pass complete — 3 SHOULD-FIX items addressed in 8266a4ac:

1. `run-qunit.mjs` no longer keeps a duplicated `serveSourceMapFile` — it now binds to the extracted `createSourceMapRouteHandler`, so future tightening propagates into the qunit harness.
2. `source-map-route.mjs` whitelist now mirrors the C# regex `^[A-Za-z0-9_-][A-Za-z0-9._-]*$` (was `includes('..' | '/' | '\')`). Parallel node smoke case covers `a:b`, `a b`, `.hidden`.
3. `source-map-browser.test.mjs` clears the 30-second startup timer on both the happy path and early exit — dropped browser test wall-time from 30.6s to 6.8s locally.

Plus `GET_PathTraversalNormalization_DoesNotServeContent` now asserts a 4xx bound (not merely `AreNotEqual(OK)`), so a hypothetical 5xx regression with no body would still fail the test.

Re-ran all affected suites post-fix:
- `SourceMap.Server.Test`: 44/44 passed
- `SourceMap.Test`: 60/60 passed
- `source-map-route.test.mjs`: 7/7 passed (+1 whitelist case)
- `source-map-browser.test.mjs`: 1/1 passed

gautam-achieveai (commented):
## [Gautam's review bot] Deep Review — PR !39 (WI #37)

### Context
This PR adds three tiers of E2E tests for the `SourceMapFileHandler` shipped in !36:
1. **C# WebApplicationFactory tests** — 10 methods driving the real ASP.NET Core pipeline
2. **C# TestHost** — Kestrel console app backing Playwright browser tests
3. **Node.js tests** — extracted `source-map-route.mjs` + smoke tests + Playwright browser test

Overall the test suite is **well-structured and thorough** — security assertions check body content (not just status codes), fixtures use the real `OwaSourceMapper.SourceMap` class, and the Playwright test gracefully degrades when dependencies are unavailable.

### Findings

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | :red_circle: HIGH | `run-qunit.mjs` | Dead import + incomplete extraction |
| 2 | :red_circle: HIGH | `TestHost/Program.cs` | `EmitMap` duplicates `FixtureMapEmitter` |
| 3 | :yellow_circle: MEDIUM | `source-map-route.mjs` | mapName validation diverges from C# handler |
| 4 | :green_circle: LOW | `source-map-browser.test.mjs` | Cleanup chain should guard individual steps |
| 5 | :green_circle: LOW | `source-map-route.mjs` | `fs.readFileSync` inconsistent with async `readFile` |
| 6 | :green_circle: LOW | `TestStartup.cs` | Static volatile field needs sequential-only comment |

See inline comments for details on each finding.

### What's Done Well
- Byte-exact assertions (`CollectionAssert.AreEqual` on raw bytes) — stronger than string matching for a file-serving handler
- `GET_SiblingPrefixRoot_Returns404_NoBodyLeak` — tests the classic containment pitfall at the full HTTP transport level
- `GET_SourceRootPlusShortName_ResolvesLikeDevTools` — reads the emitted `.map`, extracts `sourceRoot` + `sources[0]`, and composes the URL the way Chrome DevTools actually would
- `FixtureMapEmitter` uses the real compiler's `SourceMap` class — forward-compatible if the map format changes
- Playwright test `t.skip()`s gracefully when deps are missing

### Verdict: REQUEST CHANGES
Two HIGH findings need resolution before merge. The rest are improvement suggestions.

Test/Framework/TestWebApplication/run-qunit.mjs
```diff
@@ -3,6 +3,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 import { chromium } from 'playwright';
 import { fileURLToPath } from 'node:url';
+import { createSourceMapRouteHandler } from './source-map-route.mjs';
```
gautam-achieveai, line 6: **[Gautam's review bot]** :red_circle: **HIGH — Incomplete extraction: dead import, inline copy still active**

`createSourceMapRouteHandler` is imported here but **never used**. The original inline `serveSourceMapFile` function (lines 41–94) is still the one called at line 167.

1. The extraction to `source-map-route.mjs` was intended but not completed
2. There are now **two copies** of the route handler in the Node.js layer
3. The extracted module's tests validate behavior that isn't actually running in the QUnit dev server

**Fix:** Wire in the extracted handler, delete the inline copy (lines 41–94):
`js
const sourceMapHandler = createSourceMapRouteHandler(GENERATED_SCRIPTS_DIR);
// then in http.createServer: sourceMapHandler(req, res);
`
"@
    },
    @{
      path = "Test/Compiler/SourceMap.Server.TestHost/Program.cs"
      line = 189
      side = "RIGHT"
      body = @"
**[Gautam's review bot]** :red_circle: **HIGH — Duplicated EmitMap reimplements FixtureMapEmitter.Emit**

This private `EmitMap` method is structurally identical to `FixtureMapEmitter.Emit` — same `new SourceMap`, same `AddMapping(0, i*8, 0, 0)` anchor, same `map.Write()` + companion `.js`. The `FixtureMapEmitter` doc comment explicitly says it was kept MSTest-free so it can be reused from the TestHost. The two have already diverged subtly (short-name tracking, JS body content).

**Fix:** Link the emitter source file into TestHost:
`xml
<Compile Include="..\SourceMap.Server.Test\FixtureMapEmitter.cs" Link="FixtureMapEmitter.cs" />
`
Then replace `EmitMap` call with `FixtureMapEmitter.Emit(DefaultMapName, mapsDir, fixtures, sourceRoot: PathPrefix + "/" + DefaultMapName, emitLegacyAshxHandler: false)`.

Test/Framework/TestWebApplication/source-map-route.mjs
```diff
@@ -0,0 +1,85 @@
...
+//
+// Factored out of run-qunit.mjs so the behavior is testable in isolation (see
+// source-map-route.test.mjs) and reusable from any node http server.
+
+import fs from 'node:fs';
+import path from 'node:path';
+
+// Map names must not contain separators or relative segments; matches the
+// whitelist enforced by the C# handler's regex at SourceMapFileHandler.cs.
+const DISALLOWED_MAP_NAME_FRAGMENTS = ['..', '/', '\\'];
```
gautam-achieveai: **[Gautam's review bot]** :yellow_circle: **MEDIUM — mapName validation diverges from C# handler**

C# handler uses a **whitelist** regex: `^[A-Za-z0-9_-][A-Za-z0-9._-]*$`
This JS handler uses a **blacklist**: `['..', '/', '\\']`

The blacklist is more permissive — it accepts `a:b`, `a b`, `.hidden` which the C# handler rejects with 400. A developer testing with the Node handler would see 200s that become 400s in production.

**Suggestion:** Align with the C# whitelist:
`js
const MAP_NAME_PATTERN = /^[A-Za-z0-9_-][A-Za-z0-9._-]*$/;
if (!MAP_NAME_PATTERN.test(mapName)) { res.writeHead(400); res.end('Bad Request'); return true; }
`
"@
    },
    @{
      path = "Test/Framework/TestWebApplication/source-map-browser.test.mjs"
      line = 196
      side = "RIGHT"
      body = @"
**[Gautam's review bot]** :green_circle: **LOW — Cleanup chain: if browser.close() throws, server and temp dir leak**

The three cleanup steps are sequential. If `browser.close()` throws, `stopServer()` and `rmSync` are skipped. The C# `EndToEndServerTests.Cleanup` already uses the defensive pattern for `Directory.Delete`. Consider wrapping each step:
`js
if (browser) { try { await browser.close(); } catch { /* best-effort */ } }
if (serverHandle) { try { await stopServer(serverHandle.proc); } catch { /* best-effort */ } }
try { fs.rmSync(workDir, { recursive: true, force: true }); } catch { /* best-effort */ }
`
"@
    },
    @{
      path = "Test/Framework/TestWebApplication/source-map-route.mjs"
      line = 52
      side = "RIGHT"
      body = @"
**[Gautam's review bot]** :green_circle: **LOW — Inconsistent sync/async I/O**

`fs.readFileSync` here blocks the event loop on every request, but line 74 correctly uses async `fs.readFile` for the source file. When DevTools opens Sources panel, 10–50 requests may fire simultaneously, each re-reading and re-parsing the same `.map` synchronously.

For a dev-only handler this is fine in practice, but for consistency consider `fs.promises.readFile` or a simple parse cache.

Test/Compiler/SourceMap.Server.Test/TestStartup.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// <remarks>
+    /// Each test swaps its own <see cref="SourceMapFileHandlerOptions"/> onto
+    /// <see cref="CurrentOptions"/> in <c>TestInitialize</c>, then constructs a
+    /// fresh <see cref="Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory{TEntryPoint}"/>.
+    /// <c>Main</c> pulls the latched options at request time, so each factory
+    /// instance sees the options its test configured.
+    /// </remarks>
+    public class TestStartup
+    {
+        private static SourceMapFileHandlerOptions currentOptions;
```
gautam-achieveai, line 31: **[Gautam's review bot]** :green_circle: **LOW — Static volatile field is safe today but fragile**

`Volatile.Read/Write` ensures memory visibility but doesn't prevent interleaving of the set-then-construct sequence if tests run in parallel. This is safe today because MSTest defaults to sequential within a `[TestClass]`.

Suggestion: Add `[DoNotParallelize]` on `EndToEndServerTests` or a comment clarifying the sequential-only assumption.

gautam-achieveai (commented):
## [Gautam's review bot] Incremental Review — PR !39 Iteration 2

### Iteration 2 Fixes — Verified ✅

The self-review in 8266a4a addressed **3 of the 6 findings** from iteration 1. All three fixes look correct:

| Finding | Status | Notes |
|---------|--------|-------|
| HIGH #1 — Dead import in `run-qunit.mjs` | ✅ Fixed | Inline `serveSourceMapFile` deleted, now delegates to `createSourceMapRouteHandler`. Clean. |
| MEDIUM #3 — mapName blacklist vs whitelist | ✅ Fixed | `MAP_NAME_ALLOWED` regex now mirrors C# handler exactly. New test covers `a:b`, `a b`, `.hidden`. |
| Traversal assertion | ✅ Improved | Now asserts 4xx range instead of just "not 200" — catches a 500-with-no-body regression. |

Additionally, the `startServer` settle/timer refactor is a good improvement (dropped browser test wall-time from 30s to ~7s).

### Still Outstanding — 1 HIGH

| 2 | 🔴 HIGH | `TestHost/Program.cs` `EmitMap` duplicates `FixtureMapEmitter.Emit` | ❌ Not addressed |
| 4 | 🟢 LOW | Cleanup chain fragility in browser test | ❌ Not addressed (acceptable) |
| 5 | 🟢 LOW | `fs.readFileSync` inconsistency | ❌ Not addressed (acceptable) |
| 6 | 🟢 LOW | Static volatile sequential-only comment | ❌ Not addressed (acceptable) |

The LOW items are reasonable to defer. The HIGH #2 duplication is the remaining blocker — see inline comment.

### Context
- **Work Item:** #37 — E2E test for new ASP.NET Core SourceMap.Server handler
- **Parent Initiative:** #12 — Source-mapping
- **Predecessor:** !36 (the handler itself)

### Verdict: REQUEST CHANGES
One HIGH finding remains. The `EmitMap` duplication in TestHost needs to be resolved before merge — either by linking to `FixtureMapEmitter` or by explicitly justifying why the copy is intentional.

Test/Compiler/SourceMap.Server.TestHost/Program.cs
```diff
@@ -0,0 +1,221 @@
...
+
+            // Block until parent signals exit via stdin close (node harness
+            // invariant) — clean shutdown on Ctrl+C / SIGTERM is handled by
+            // ASP.NET Core's lifetime.
+            Console.In.ReadToEnd();
+            app.StopAsync().GetAwaiter().GetResult();
+            return 0;
+        }
+
+        private static void EmitMap(
```
gautam-achieveai: **[Gautam's review bot]** :red_circle: **HIGH — Still unresolved: `EmitMap` duplicates `FixtureMapEmitter.Emit`**

This private `EmitMap` (lines 189-219) is structurally identical to `FixtureMapEmitter.Emit` — same `new SourceMap`, same `AddMapping(0, i*8, 0, 0)` anchor, same `map.Write()` + companion `.js`. The two have already diverged:

- `FixtureMapEmitter` has argument validation, short-name tracking via `ToShortName()`, and `BuildFixtureJs()`
- This copy has none of that

The `FixtureMapEmitter` doc comment (line 22-23) explicitly says:
> *Kept deliberately free of any MSTest dependency so it can be reused from the SourceMap.Server.TestHost console entry point*

**Fix:** Link the emitter source into TestHost's `.csproj`:
`xml
<Compile Include="..\..\SourceMap.Server.Test\FixtureMapEmitter.cs" Link="FixtureMapEmitter.cs" />
`
Then replace the `EmitMap` call at line 111 with:
`csharp
FixtureMapEmitter.Emit(DefaultMapName, mapsDir, fixtures,
    sourceRoot: PathPrefix + "/" + DefaultMapName,
    emitLegacyAshxHandler: false);
`

_(Note: This finding was originally posted in iteration 1 but was embedded inside another comment's body due to a formatting issue. Reposting as a standalone comment for clarity.)_

gautam-achieveai: [Gautam's Dev bot]

Fixed in 69dff7d1.

- Linked `FixtureMapEmitter.cs` into `SourceMap.Server.TestHost.csproj` via `<Compile Link=...>` — exactly as suggested.
- Replaced the private `EmitMap` with `FixtureMapEmitter.Emit(DefaultMapName, mapsDir, fixtures, sourceRoot: PathPrefix + "/" + DefaultMapName, emitLegacyAshxHandler: false)`.
- `DefaultMapName` now aliases `FixtureMapEmitter.DefaultMapName` so the two projects can't drift.
- Removed the duplicated 31-line `EmitMap` method entirely.

Browser and WAF suites now emit byte-identical `.map` + `.js` pairs from a single code path. Build clean, `SourceMap.Server.Test` 44/44 pass.

gautam-achieveai (commented):
**[Gautam's review bot]** :white_check_mark: **Iteration 3 Review — APPROVE**

The remaining HIGH finding (EmitMap duplication in TestHost) is cleanly resolved in `69dff7d`:

- `FixtureMapEmitter.cs` shared via `<Compile Link>` — correct approach for MSTest-free source sharing
- `DefaultMapName` aliases `FixtureMapEmitter.DefaultMapName` — prevents future drift
- `emitted.MapPath` used instead of manual path concat — single source of truth
- 31-line duplicate deleted

**All findings status:**
| Sev | Finding | Status |
|-----|---------|--------|
| :red_circle: HIGH | `run-qunit.mjs` dead import | :white_check_mark: Fixed iter 2 |
| :red_circle: HIGH | EmitMap duplication | :white_check_mark: Fixed iter 3 |
| :yellow_circle: MED | mapName validation divergence | :white_check_mark: Fixed iter 2 |
| :green_circle: LOW | Cleanup chain guards | :white_circle: Acceptable |
| :green_circle: LOW | sync/async I/O consistency | :white_circle: Acceptable |
| :green_circle: LOW | Static volatile comment | :white_circle: Acceptable |

All HIGH and MEDIUM findings resolved. Remaining LOWs are acceptable for test-only code. **LGTM.**

Test/Compiler/SourceMap.Server.TestHost/Program.cs
gautam-achieveai: **[Gautam's review bot]** :white_check_mark: Verified in `69dff7d`. Clean fix — `<Compile Link>` shares the source correctly, `DefaultMapName` aliased, 31-line duplicate removed. LGTM.

PR #40, opened 2026-04-24
WI-34: WebStoreTable explicit pagination/streaming API

### LINKED ISSUE #34: WebStoreTable: replace 1<<20 default row cap with explicit pagination/streaming API
## Context

Follow-up from PR #31 review ([thread](https://github.com/achieveai/NScript/pull/31#discussion_r3114601539)).

`WebStoreTable.GetCursor` and `Query(Query.All)` currently default their row cap to `1 << 20` (~1M). The magic number was inherited from the MCQdb port. It hides three real problems:

1. It is not a true `All` semantic — a dataset larger than 1M silently truncates.
2. It forces every caller to materialize the full result set into a `List<T>` in memory.
3. It makes it hard to spot accidental full-table scans because the cap is invisible at the call site.

## Proposed work

- Remove the implicit `1 << 20` default.
- Make `Query.All` require an explicit `Limit` or a streaming `ForEach` overload.
- Audit existing callers (`TodoDataService.GetAllTodos`, `TodoDataService.GetAllFolders`, and any internal telemetry store use) and pick an appropriate cap / streaming shape per caller.
- Surface large unbounded queries at build time (analyzer or compile-time attribute) if feasible.

## Out of scope for PR #31

PR #31 is scoped to landing the typed WebStore layer on top of IndexedDB. Changing the default requires an API review across all current callers and is best done as a focused follow-up.

gautam-achieveai:
[Gautam's Dev bot]

## Summary

Replaces `WebStoreTable`'s silent `1 << 20` row cap with an explicit pagination/streaming API. Callers must now opt into either streaming via `ForEach` or a bounded materialising `Query` with `QueryBuilder.Limit(n)`.

Fixes #34.

## Motivation

The old read paths silently truncated at ~1M rows — a table that grew past the cap would quietly return partial results with no error. There was also no streaming alternative, so callers had to materialise every match into a `List<T>`.

## Key Decisions

- **Rejected unbounded materialisation**: `Query(Query.All)` without a filter now rejects with a descriptive error instead of resolving. The message names `Query.All` and points callers at `ForEach` or `QueryBuilder.Limit(n)`. Callers with a filter predicate are still allowed to use `Query.All` because the predicate bounds the materialised set.
- **Added `ForEach` streaming API**: `Promise<int> ForEach(Query, Func<TValue,bool>, WebStoreTransaction)`. Visitor returns `false` to stop early; resolve value is the number of records visited.
- **Strict `Limit` semantics**: `CursorIterator` honours `query.Limit` exactly — `null` = unbounded, `0` = no visits, `N` = exactly N. The cap check is placed before the visit so `Limit(0)` is a true no-op (previously would have visited one row due to the decrement-then-check order).
- **Symmetric Promise-based error surface**: Both `ForEach(null)` and `Query(Query.All)` report argument errors through the Promise rejection path so callers using `.Then(_, onRejected)` get a uniform API.
- **Call-site visible cap in TodoApp**: `TodoDataService.GetAllTodos`/`GetAllFolders` now pass `QueryBuilder.Limit(TodoLimits.MaxRecords)` where `TodoLimits.MaxRecords = 10000`. The bound is visible at the call site rather than buried inside the cursor iterator.

## Test Plan

- [x] Debug build (`dotnet build NScript_Full.sln -c Debug`) green
- [x] All QUnit suites green: 71 Framework + 93 UI + 26 TodoApp + 19 Data = **209/209 passed**
- [x] New Data.Test cases:
  - `TestForEachAllVisitsEveryRecord` — streaming hits every record
  - `TestQueryAllMaterialisingThrows` — rejects with descriptive message
  - `TestForEachStopsEarly` — visitor `false` stops iteration
  - `TestLimitedQueryRespectsExplicitLimit` — `Limit == recordCount`
  - `TestLimitSmallerThanRecordCountTruncates` — `Limit(2)` on 5 rows
  - `TestLimitZeroVisitsNothing` — direct `Limit(0)` no-op test
  - `TestForEachEmptyTableVisitsNothing` — empty-store short-circuit
  - `TestForEachNullVisitRejects` — argument validation via rejection
- [x] Sunlight.Framework.Data.Test wired into `run-qunit.mjs` so the suite runs by default in CI
- [x] Two self-review iterations via `code-reviewer:pr-review` (NScript compliance + test coverage), all flagged findings addressed

### DIFF
--- a/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
+++ b/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
@@ -5,6 +5,7 @@ namespace Sunlight.Framework.Data.WebStore
     using System.Runtime.CompilerServices;
     using System.Web.Html;
     using System.Web.Html.Data.IndexedDB;
+    using QueryDef = Sunlight.Framework.Data.WebStore.Query;
 
     internal enum CursorOrCountRequestMode
     {
@@ -173,6 +174,14 @@ namespace Sunlight.Framework.Data.WebStore
         /// <summary>
         /// Runs <paramref name="query"/> and returns all matching records
         /// (filtered by <paramref name="inclFilter"/> when supplied).
+        /// <para>
+        /// The unbounded <see cref="WebStore.Query.All"/> singleton is rejected on
+        /// this materialising path — it would buffer the entire table into a
+        /// <see cref="List{T}"/> with no row ceiling. Use
+        /// <see cref="ForEach(Query, Func{TValue, bool}, WebStoreTransaction)"/>
+        /// to stream the full table, or build a query with an explicit
+        /// <c>QueryBuilder.Limit(n)</c> to bound the result set.
+        /// </para>
         /// </summary>
         public Promise<List<TValue>> Query(
             Query query,
@@ -187,6 +196,30 @@ namespace Sunlight.Framework.Data.WebStore
                     resolve,
                     reject,
                     false));
+        }
+
+        /// <summary>
+        /// Streams every record matching <paramref name="query"/> through
+        /// <paramref name="visit"/> without materialising the result set. The
+        /// visitor returns <c>true</c> to continue, <c>false</c> to stop early.
+        /// Resolves with the number of records visited (including the one the
+        /// visitor returned <c>false</c> for, if any).
+        /// </summary>
+        public Promise<int> ForEach(
+            Query query,
+            Func<TValue, bool> visit,
+            WebStoreTransaction transaction = null)
+        {
+            if (visit == null)
+            { throw new Exception("visit can't be null"); }
+
+            return new Promise<int>((resolve, reject) =>
+                this.ForEachInternal(
+                    transaction,
+                    query,
+                    visit,
+                    resolve,
+                    reject));
         }
 
         /// <summary>Same as <see cref="Query"/> but returns primary keys only (key cursor).</summary>
@@ -475,7 +508,7 @@ namespace Sunlight.Framework.Data.WebStore
             { return; }
 
             var skip = query.Skip ?? 0;
-            var limit = query.Limit ?? 1 << 20;
+            int? remaining = query.Limit;
             bool firstLoop = skip > 0;
 
             request.OnSuccess += (req, evt) =>
@@ -499,10 +532,14 @@ namespace Sunlight.Framework.Data.WebStore
                     if (!onIterate(cursor))
                     { return; }
 
-                    if (--limit == 0)
+                    if (remaining.HasValue)
                     {
-                        _ = onIterate(null);
-                        return;
+                        remaining = remaining.Value - 1;
+                        if (remaining.Value == 0)
+                        {
+                            _ = onIterate(null);
+                            return;
+                        }
                     }
                 }
 
@@ -697,6 +734,43 @@ namespace Sunlight.Framework.Data.WebStore
                 reject);
         }
 
+        private void ForEachInternal(
+            WebStoreTransaction transaction,
+            Query query,
+            Func<TValue, bool> visit,
+            Action<int> resolve,
+            Action<object> reject)
+        {
+            transaction = this.TransactionOrDefault(
+                transaction,
+                TransactionKind.Read);
+
+            int count = 0;
+            this.CursorIterator(
+                transaction,
+                query,
+                null,
+                (cursor) =>
+                {
+                    if (cursor == null)
+                    {
+                        resolve(count);
+                        return true;
+                    }
+
+                    count = count + 1;
+                    if (!visit(cursor.Value))
+                    {
+                        resolve(count);
+                        return false;
+                    }
+
+                    return true;
+                },
+                reject,
+                false);
+        }
+
         private void QueryInternal<U>(
             WebStoreTransaction transaction,
             Query query,
@@ -705,6 +779,15 @@ namespace Sunlight.Framework.Data.WebStore
             Action<object> reject,
             bool isKeyQuery)
         {
+            if (Object.ReferenceEquals(query, QueryDef.All) && filter == null)
+            {
+                reject(new Exception(
+                    "Query(Query.All) is disallowed because it can materialise an "
+                    + "unbounded result set into memory. Use ForEach for streaming, "
+                    + "or set an explicit cap via QueryBuilder.Limit(n)."));
+                return;
+            }
+
             transaction = this.TransactionOrDefault(
                 transaction,
                 TransactionKind.Read);
--- a/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
+++ b/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
@@ -789,7 +789,7 @@ namespace Sunlight.Framework.Data.WebStore
             Action<object> reject,
             bool isKeyQuery)
         {
-            if (Object.ReferenceEquals(query, QueryDef.All) && filter == null)
+            if (Object.ReferenceEquals(query, QueryDef.All) && filter == null && !isKeyQuery)
             {
                 reject(new Exception(
                     "Query(Query.All) is disallowed because it can materialise an "
--- a/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
+++ b/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
@@ -210,16 +210,20 @@ namespace Sunlight.Framework.Data.WebStore
             Func<TValue, bool> visit,
             WebStoreTransaction transaction = null)
         {
-            if (visit == null)
-            { throw new Exception("visit can't be null"); }
-
             return new Promise<int>((resolve, reject) =>
+            {
+                if (visit == null)
+                {
+                    reject(new Exception("visit can't be null"));
+                    return;
+                }
                 this.ForEachInternal(
                     transaction,
                     query,
                     visit,
                     resolve,
-                    reject));
+                    reject);
+            });
         }
 
         /// <summary>Same as <see cref="Query"/> but returns primary keys only (key cursor).</summary>
@@ -529,6 +533,12 @@ namespace Sunlight.Framework.Data.WebStore
 
                 if (filter == null || filter(cursor.Value))
                 {
+                    if (remaining.HasValue && remaining.Value <= 0)
+                    {
+                        _ = onIterate(null);
+                        return;
+                    }
+
                     if (!onIterate(cursor))
                     { return; }
 
--- a/Test/Framework/Sunlight.Framework.Data.Test/WebStoreCrudTests.cs
+++ b/Test/Framework/Sunlight.Framework.Data.Test/WebStoreCrudTests.cs
@@ -479,6 +479,135 @@ namespace Sunlight.Framework.Data.Test
         }
 
         /// <summary>
+        /// <c>ForEach</c> combined with <c>QueryBuilder.Limit(2)</c> on a
+        /// 5-record store must stop after 2 records — invoking the visitor
+        /// exactly twice. Pins the interaction between the streaming ForEach
+        /// path and the explicit Limit cap, which were introduced together but
+        /// previously only tested independently.
+        /// </summary>
+        [Test]
+        public static void TestForEachRespectsExplicitLimit(Assert assert)
+        {
+            var done = assert.Async(1);
+            var factory = new WebStoreFactory();
+            var dbName = NewDbName();
+
+            factory.Create(BuildSchema(dbName)).Then<bool>(delegate(WebStoreClient client)
+            {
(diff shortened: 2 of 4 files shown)


Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
```diff
@@ -705,6 +789,15 @@ private void QueryInternal<U>(
             Action<object> reject,
             bool isKeyQuery)
         {
+            if (Object.ReferenceEquals(query, QueryDef.All) && filter == null)
```
gautam-achieveai: **[Gautam's review bot]** :red_circle: **HIGH - Bug: `QueryKeys(Query.All)` silently broken**

`QueryKeys()` at L234-241 calls `QueryInternal<TKey>(transaction, query, null, ...)` - always passing `null` as the filter. Combined with this guard, `QueryKeys(Query.All)` now **always rejects**.

Before this PR, `QueryKeys(Query.All)` worked. This is a behavioral regression.

The error message says *"Use ForEach for streaming"* but there's no `ForEachKeys` equivalent, so `QueryKeys(Query.All)` callers have no migration path.

The `isKeyQuery` parameter is already available at this call site (line 790). Fix:
```csharp
if (Object.ReferenceEquals(query, QueryDef.All) && filter == null && !isKeyQuery)
```

Alternatively, if rejecting `QueryKeys(Query.All)` *is* intentional, add a test that pins the rejection and update the error message.

Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
```diff
@@ -697,6 +744,43 @@ private void QueryDeleteInternal(
...
+                (cursor) =>
+                {
+                    if (cursor == null)
+                    {
+                        resolve(count);
+                        return true;
+                    }
+
+                    count = count + 1;
+                    if (!visit(cursor.Value))
```
gautam-achieveai, line 772: **[Gautam's review bot]** :yellow_circle: **HIGH - Pre-existing risk now public: unprotected `visit()` callback**

If `visit(cursor.Value)` throws, the exception escapes the IDB `OnSuccess` handler. The Promise never resolves or rejects - callers hang indefinitely.

Pre-existing pattern but `ForEach` is the first public API that invites **user-provided** callbacks into the cursor loop.

Non-blocking (predates this PR). Recommend tracked follow-up to wrap the `visit` call:
```csharp
try { if (!visit(cursor.Value)) { resolve(count); return false; } }
catch (Exception ex) { reject(ex); return false; }
```

Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
```diff
@@ -705,6 +789,15 @@ private void QueryInternal<U>(
             Action<object> reject,
             bool isKeyQuery)
         {
+            if (Object.ReferenceEquals(query, QueryDef.All) && filter == null)
+            {
+                reject(new Exception(
+                    "Query(Query.All) is disallowed because it can materialise an "
```
gautam-achieveai, line 795: **[Gautam's review bot]** :speech_balloon: **MEDIUM - Design note: `ReferenceEquals` is identity-only**

This guard only catches the exact `Query.All` singleton. Bypass possible via public `Query` constructor, though `QueryBuilder.Build()` already throws for unconstrained case.

For defense-in-depth, consider structural check (`query.Range == null && query.Limit == null && ...`) or making `Query` constructor `internal`.

Non-blocking - current guard catches the common case.

Test/Framework/Sunlight.Framework.Data.Test/WebStoreCrudTests.cs
```diff
@@ -182,6 +190,294 @@ public static void TestQueryAllReturnsEveryRecord(Assert assert)
...
+                    done();
+                    return true;
+                });
+                return true;
+            });
+        }
+
         /// <summary>
         /// Defect #3 regression — with the missing [JsonType] on IDBIndexParameters,
         /// the <c>Unique</c> field name would not reach the native IDB call so
```
gautam-achieveai, line 612: **[Gautam's review bot]** :red_circle: **HIGH - Missing test: `ForEach` + `Limit` combination**

All `ForEach` tests use `Query.All` (Limit=null). All `Limit` tests use `Query()` (materializing path). The two flagship features are tested independently but never together.

Suggested: Seed 5 records, `ForEach` with `Limit(2)`, assert count=2 and visitor called exactly twice.

Test/Framework/Sunlight.Framework.Data.Test/WebStoreCrudTests.cs
```diff
@@ -182,6 +190,294 @@ public static void TestQueryAllReturnsEveryRecord(Assert assert)
...
+                                message != null && message.IndexOf("Query.All") >= 0,
+                                "rejection message names Query.All so callers can diagnose");
+                            assert.IsTrue(
+                                message.IndexOf("Limit") >= 0 || message.IndexOf("ForEach") >= 0,
+                                "rejection message points callers at Limit or ForEach as the fix");
+                            client.Close();
+                            done();
+                            return true;
+                        });
+                    return true;
```
gautam-achieveai, line 233: **[Gautam's review bot]** :yellow_circle: **HIGH - Missing test: `Query(Query.All, filter)` positive path**

The guard has intentional `&& filter == null` escape hatch. No test pins the positive path where `Query(Query.All, someFilter)` succeeds.

Without it, refactoring the guard to just `ReferenceEquals(query, QueryDef.All)` would break filtered scans with no test failure.

Suggested: Seed records with mixed categories, `Query(Query.All, row => row.Category == "todo")` asserts it resolves with filtered results.

gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of PR !40 (linked to #34).

## Verdict: REQUEST CHANGES

### Summary
Solid design overall — extracting `ForEach` streaming API, sharing `CursorIterator`, and migrating `TodoDataService` to explicit `Limit()` are clean choices. Performance review found zero concerns. Two blocking issues need attention before merge.

### Blocking Issues

| # | Severity | Finding | Fix |
|---|----------|---------|-----|
| F1 | **HIGH** | `QueryKeys(Query.All)` silently broken — the guard in `QueryInternal` rejects it since `QueryKeys` passes `filter=null`. No `ForEachKeys` alternative exists. | Add `&& !isKeyQuery` to guard at L792 |
| F4 | **HIGH** | `ForEach` + `Limit` combination untested — the PR's two flagship features are tested independently but never together. | Add one `ForEach(Limit(2))` test with 5 records |

### Non-Blocking Findings

| # | Severity | Finding |
|---|----------|---------|
| F2 | HIGH | Unprotected `visit()` callback — if visitor throws in `CursorIterator.OnSuccess`, Promise hangs forever. Pre-existing pattern but `ForEach` makes it public-facing. Recommend tracked follow-up. |
| F3 | HIGH | `Query(Query.All, inclFilter)` bypass path untested — the `&& filter == null` escape hatch has no positive test. |
| F5 | MEDIUM | `ReferenceEquals` guard is identity-only — bypassable via public `Query` constructor. Pragmatic speed-bump. |
| F6 | MEDIUM | `Delete`/`Update` with `Query.All` not guarded — asymmetry with read paths. Pre-existing. |
| F7 | MEDIUM | Inconsistent null-argument validation — `ForEach` correctly uses `reject()`, existing `Update` uses synchronous `throw`. |

### Review Agents Used
architecture-review, performance-review, exception-handling-review, test-coverage-review, review-grader

### What's Good
- Clean separation: `ForEach` → `ForEachInternal` → `CursorIterator` follows existing delegation pattern
- `TodoLimits` constant makes the bound visible at call sites — much better than the hidden `1 << 20`
- Excellent test suite — 8 new tests with clear naming and doc comments explaining *why* each exists
- Correct `reject()` pattern in `ForEach` null-visit guard (sets good precedent)
- `AllWithLimit()` helper is appropriately simple for its single-consumer scope

Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 8d63a97b. Guard now reads `ReferenceEquals(query, QueryDef.All) && filter == null && !isKeyQuery`. `QueryKeys` passes `isKeyQuery=true` at L791 (now), so `QueryKeys(Query.All)` resolves normally — key-only scans carry no materialisation risk. Added `TestQueryKeysAllResolves` as regression pin.

Test/Framework/Sunlight.Framework.Data.Test/WebStoreCrudTests.cs
gautam-achieveai: [Gautam's Dev bot]

Added `TestForEachRespectsExplicitLimit` in 8d63a97b. Seeds 5 records, calls `ForEach` with a `Limit(2)` query, asserts visit count == 2 and visitor invoked exactly twice.

Test/Framework/Sunlight.Framework.Data.Test/WebStoreCrudTests.cs
gautam-achieveai: [Gautam's Dev bot]

Added `TestQueryAllWithFilterResolvesFiltered` in 8d63a97b. Seeds 3 records with mixed categories, calls `Query(Query.All, row => row.Category == "todo")`, asserts 2 filtered rows returned. Pins the positive filter-escape-hatch so refactoring the guard can't silently break filtered scans.

Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
gautam-achieveai: [Gautam's Dev bot]

Acknowledged — this is a pre-existing pattern across every cursor consumer, not specific to ForEach. Tracking as follow-up issue rather than bundling into this PR to keep the change surgical. Will open a separate issue to wrap `visit` in try/catch → `reject(ex)` and propagate the same guard to the other cursor call sites.

Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
gautam-achieveai: [Gautam's Dev bot]

Noted — leaving `ReferenceEquals` as the speed-bump for now. `QueryBuilder.Build()` already rejects unconstrained `Query()` construction, and the only documented path to reach the guard is via the `Query.All` singleton. If a structural bypass shows up in practice we'll tighten; happy to revisit as a follow-up if you'd prefer a hard guard.

gautam-achieveai:
[Gautam's Dev bot]

Addressed review findings in 8d63a97b.

**Blocking fixes**
- **F1** — Guard in `QueryInternal` now reads `ReferenceEquals(query, QueryDef.All) && filter == null && !isKeyQuery`. `QueryKeys(Query.All)` resolves normally again. Added `TestQueryKeysAllResolves` regression pin.
- **F4** — Added `TestForEachRespectsExplicitLimit` covering the `ForEach` + `Limit(2)` interaction on a 5-record store.

**Non-blocking addressed**
- **F3** — Added `TestQueryAllWithFilterResolvesFiltered` to pin the positive `filter != null` escape-hatch path.
- **F2** — Tracked as #41 (WebStore cursor: wrap visitor callbacks in try/catch → reject). Agreeing this is real, but pre-existing across all cursor consumers — keeping the current PR surgical.
- **F5, F6, F7** — Design-note / pre-existing / non-blocking. Happy to address as follow-up if preferred; current behaviour is explicit and documented via tests.

**Quality gates**
- Debug build: pass.
- QUnit: 212/212 pass (71 + 93 + 26 + 22 Data).

gautam-achieveai (commented):
**[Gautam's review bot]** :white_check_mark: **APPROVE** — Incremental review of iteration 4 (`8d63a97b`)

All blocking findings from the previous review are resolved:

| Finding | Status | Verification |
|---------|--------|-------------|
| **F1** (HIGH, blocker) — QueryKeys regression | :white_check_mark: Fixed | Guard adds `&& !isKeyQuery`. QueryKeys passes `true` at L241. `TestQueryKeysAllResolves` pins it. |
| **F4** (HIGH, blocker) — ForEach+Limit untested | :white_check_mark: Fixed | `TestForEachRespectsExplicitLimit`: 5 records, Limit(2), asserts count==2 and visits==2. |
| **F3** (HIGH, non-blocking) — filter escape hatch untested | :white_check_mark: Fixed | `TestQueryAllWithFilterResolvesFiltered` pins the positive path. |
| **F2** (HIGH, non-blocking) — visit() callback safety | :hourglass_flowing_sand: Tracked | Pre-existing across all cursor consumers; appropriate to defer. |

**Code quality:** The one-line guard fix is surgical and correct. All 3 new tests follow existing patterns with good assertions. No new issues introduced.

**Verdict: APPROVE** (posted as COMMENT due to same-user limitation)

**Work item:** #34

PR #43, opened 2026-04-24
WI-41: wrap WebStore cursor visitor callbacks in try/catch → reject

### LINKED ISSUE #41: WebStore cursor: wrap visitor callbacks in try/catch → reject
[Gautam's Dev bot]

Follow-up from PR #40 review (finding F2).

## Problem

In `WebStoreTable.CursorIterator` consumers, user-provided callbacks (e.g. the `visit` delegate passed to `ForEach`) run inside an IDB `OnSuccess` handler. If the callback throws, the exception escapes the event handler and the wrapping `Promise` never resolves or rejects — callers hang forever.

This is a pre-existing pattern across every cursor consumer (`Query`, `QueryKeys`, `ForEach`, `QueryDelete`), not specific to `ForEach`. But `ForEach` is the first public API that invites user-provided callbacks into the cursor loop, so the blast radius is now bigger.

## Proposed fix

Wrap every callback invocation in a try/catch that routes exceptions to the `reject` continuation and aborts the cursor. Apply consistently to all cursor-iterator call sites:

```csharp
try
{
    if (!visit(cursor.Value))
    {
        resolve(count);
        return false;
    }
}
catch (Exception ex)
{
    reject(ex);
    return false;
}
```

## Scope

- `Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs`: `ForEachInternal`, `QueryInternal` (filter callback), `QueryDeleteInternal` (filter callback).
- Test: seed records, pass a `visit`/`filter` that throws, assert Promise rejects with the thrown error.

## Related

- PR #40 (WI-34): introduced `ForEach` streaming API.
- Review finding F2.

gautam-achieveai:
[Gautam's Dev bot]

## Summary

Closes #41.

`WebStoreTable.CursorIterator` invokes three user-provided callback seams (`filter`, `visit`, `updateFunc`) inside an IDB `OnSuccess` handler. If any callback threw, the exception escaped the event handler and the wrapping `Promise` never settled — callers hung forever.

This PR centralises a try/catch barrier inside `CursorIterator` that covers every user-callback seam. On throw, the barrier rejects the returned `Promise` with the **original** exception and marks the iterator `aborted` so subsequent `OnSuccess`/`OnError` events are discarded.

## Why one barrier instead of three per-caller try/catches

All three seams funnel through `CursorIterator.OnSuccess`:
- `filter(cursor.Value)` — direct call inside `CursorIterator`.
- `visit(cursor.Value)` — inside `ForEachInternal`'s `onIterate` lambda, reached via `onIterate(cursor)`.
- `updateFunc(cursor.Value)` — inside `QueryUpdateOrDeleteInternal`'s `onIterate` lambda, reached via `onIterate(cursor)`.

Wrapping each `onIterate(cursor)` call edge transitively catches `visit()` and `updateFunc()` because they execute synchronously inside the delegate. One barrier; any new `*Internal` consumer is covered automatically.

## On-throw behaviour

- Call `reject(ex)` with the thrown exception (preserved, not wrapped).
- Do **not** continue the cursor → IDB auto-closes once `OnSuccess` returns without `Continue()`.
- Do **not** `transaction.Abort()` — matches the existing cursor `OnError` path, which also does not abort. For `Delete`/`Update`, writes are batched via `BatchUpdateOrDeleteWithKeys` only after the cursor terminates, so a mid-cursor abort never reaches the write path.
- An `aborted` flag guards the catch path and the `OnError` handler so a catch-then-error race settles the promise exactly once (mirrors `AddBatchInternal.requestAborted`).

## Tests

Four new regression tests in `WebStoreCrudTests.cs`, one per user-callback seam:

| Test | Exercises |
|------|-----------|
| `TestForEachVisitThrowsRejects` | `ForEach` visitor throw → reject with original exception |
| `TestQueryFilterThrowsRejects` | `Query(query, filter)` filter throw → reject |
| `TestDeleteFilterThrowsRejects` | `Delete(query, filter)` filter throw → reject; both records survive |
| `TestUpdateFuncThrowsRejects` | `Update(query, updateFunc)` updateFunc throw → reject; both record scores unchanged |

Each test asserts the rejection propagates the **original** exception message (`boom-visit`, `boom-filter`, `boom-delete-filter`, `boom-update`) so the fix hasn't silently wrapped caller errors.

## Test plan
- [x] `dotnet build NScript_Full.sln -c Debug` green (0 errors, 188 pre-existing warnings).
- [x] Browser QUnit suite via Playwright: 26/26 Data tests pass (22 existing + 4 new).
- [x] Pre-existing regression tests unchanged (`TestForEachAllVisitsEveryRecord`, `TestForEachStopsEarly`, `TestForEachRespectsExplicitLimit`, `TestQueryAllWithFilterResolvesFiltered`, `TestQueryKeysAllResolves`, `TestLimit*`).

## Risk

Low. Pure additive safety net. Happy-path fallthrough is byte-identical to today's code — the `try` blocks only trigger on exception. The `aborted` flag gates a pre-existing `reject`-only error path that already tolerated double-settle.

### DIFF
--- a/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
+++ b/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
@@ -531,7 +531,8 @@ namespace Sunlight.Framework.Data.WebStore
 
                 if (firstLoop)
                 {
-                    cursor.Advance(skip);
+                    try { cursor.Advance(skip); }
+                    catch (Exception ex) { aborted = true; reject(ex); return; }
                     firstLoop = false;
                     return;
                 }
@@ -567,14 +568,17 @@ namespace Sunlight.Framework.Data.WebStore
                     }
                 }
 
-                cursor.Continue();
+                try { cursor.Continue(); }
+                catch (Exception ex) { aborted = true; reject(ex); }
             };
 
             var onError = HandleError(reject);
             request.OnError += (req, evt) =>
             {
+                evt.PreventDefault();
                 if (aborted)
                 { return; }
+                aborted = true;
                 onError(req, evt);
             };
         }
--- a/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
+++ b/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
@@ -572,14 +572,15 @@ namespace Sunlight.Framework.Data.WebStore
                 catch (Exception ex) { aborted = true; reject(ex); }
             };
 
-            var onError = HandleError(reject);
             request.OnError += (req, evt) =>
             {
                 evt.PreventDefault();
                 if (aborted)
                 { return; }
                 aborted = true;
-                onError(req, evt);
+                reject(new EventBasedException(
+                    req.Error.ExceptionMessage(),
+                    evt));
             };
         }
 
--- a/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
+++ b/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
@@ -514,13 +514,18 @@ namespace Sunlight.Framework.Data.WebStore
             var skip = query.Skip ?? 0;
             int? remaining = query.Limit;
             bool firstLoop = skip > 0;
+            bool aborted = false;
 
             request.OnSuccess += (req, evt) =>
             {
+                if (aborted)
+                { return; }
+
                 var cursor = Type.AS<object, IDBCursor<TKey, TValue>>(req.Result);
                 if (cursor == null)
                 {
-                    _ = onIterate(null);
+                    try { _ = onIterate(null); }
+                    catch (Exception ex) { aborted = true; reject(ex); }
                     return;
                 }
 
@@ -531,15 +536,23 @@ namespace Sunlight.Framework.Data.WebStore
                     return;
                 }
 
-                if (filter == null || filter(cursor.Value))
+                bool accept;
+                try { accept = filter == null || filter(cursor.Value); }
+                catch (Exception ex) { aborted = true; reject(ex); return; }
+
+                if (accept)
                 {
                     if (remaining.HasValue && remaining.Value <= 0)
                     {
-                        _ = onIterate(null);
+                        try { _ = onIterate(null); }
+                        catch (Exception ex) { aborted = true; reject(ex); }
                         return;
                     }
 
-                    if (!onIterate(cursor))
+                    bool cont;
+                    try { cont = onIterate(cursor); }
+                    catch (Exception ex) { aborted = true; reject(ex); return; }
+                    if (!cont)
                     { return; }
 
                     if (remaining.HasValue)
@@ -547,7 +560,8 @@ namespace Sunlight.Framework.Data.WebStore
                         remaining = remaining.Value - 1;
                         if (remaining.Value == 0)
                         {
-                            _ = onIterate(null);
+                            try { _ = onIterate(null); }
+                            catch (Exception ex) { aborted = true; reject(ex); }
                             return;
                         }
                     }
@@ -556,7 +570,13 @@ namespace Sunlight.Framework.Data.WebStore
                 cursor.Continue();
             };
 
-            request.OnError += HandleError(reject);
+            var onError = HandleError(reject);
+            request.OnError += (req, evt) =>
+            {
+                if (aborted)
+                { return; }
+                onError(req, evt);
+            };
         }
 
         private IDBRequest GetCursor(
--- a/Test/Framework/Sunlight.Framework.Data.Test/WebStoreCrudTests.cs
+++ b/Test/Framework/Sunlight.Framework.Data.Test/WebStoreCrudTests.cs
@@ -679,8 +679,10 @@ namespace Sunlight.Framework.Data.Test
                 {
                     table.UpSert(NewEntity("b", "todo", 2)).Then<bool>(delegate(string k2)
                     {
+                        int visitCount = 0;
                         table.ForEach(Query.All, delegate(CrudEntity row)
                         {
+                            visitCount++;
                             throw new Exception("boom-visit");
                         }).Then<bool, object>(
                             delegate(int count)
@@ -692,6 +694,7 @@ namespace Sunlight.Framework.Data.Test
                             },
                             delegate(object err)
                             {
+                                assert.Equal(visitCount, 1, "visitor called exactly once before abort");
                                 assert.IsTrue(err != null, "rejection has an error");
                                 var message = ((Exception)err).Message;
                                 assert.IsTrue(
--- a/Test/Framework/Sunlight.Framework.Data.Test/WebStoreCrudTests.cs
+++ b/Test/Framework/Sunlight.Framework.Data.Test/WebStoreCrudTests.cs
@@ -657,6 +657,246 @@ namespace Sunlight.Framework.Data.Test
             });
         }
 
+        /// <summary>
+        /// A <c>ForEach</c> visitor that throws must reject the returned
+        /// Promise with the original exception — not hang the caller. Before
+        /// WI-41, the throw escaped the IDB <c>OnSuccess</c> handler and the
+        /// promise was never settled. The centralised try/catch in
+        /// <c>CursorIterator</c> routes the thrown exception to <c>reject</c>.
+        /// </summary>
+        [Test]
+        public static void TestForEachVisitThrowsRejects(Assert assert)
+        {
+            var done = assert.Async(1);
+            var factory = new WebStoreFactory();
+            var dbName = NewDbName();
+
+            factory.Create(BuildSchema(dbName)).Then<bool>(delegate(WebStoreClient client)
+            {
+                var table = client.Table<string, CrudEntity>(TableName);
+
+                table.UpSert(NewEntity("a", "todo", 1)).Then<bool>(delegate(string k1)
+                {
+                    table.UpSert(NewEntity("b", "todo", 2)).Then<bool>(delegate(string k2)
+                    {
+                        table.ForEach(Query.All, delegate(CrudEntity row)
+                        {
+                            throw new Exception("boom-visit");
+                        }).Then<bool, object>(
+                            delegate(int count)
+                            {
+                                assert.IsTrue(false, "throwing visitor should reject, not resolve");
+                                client.Close();
+                                done();
+                                return true;
+                            },
+                            delegate(object err)
+                            {
+                                assert.IsTrue(err != null, "rejection has an error");
+                                var message = ((Exception)err).Message;
+                                assert.IsTrue(
+                                    message != null && message.IndexOf("boom-visit") >= 0,
+                                    "original thrown exception propagates to reject");
+                                client.Close();
+                                done();
+                                return true;
+                            });
+                        return true;
+                    });
+                    return true;
+                });
+                return true;
+            });
+        }
+
+        /// <summary>
+        /// A <c>Query</c> filter that throws must reject the returned Promise
+        /// with the original exception. Exercises the <c>filter(cursor.Value)</c>
+        /// call edge in <c>CursorIterator</c>, distinct from the <c>onIterate</c>
+        /// edge covered by <c>TestForEachVisitThrowsRejects</c>.
+        /// </summary>
+        [Test]
+        public static void TestQueryFilterThrowsRejects(Assert assert)
+        {
+            var done = assert.Async(1);
+            var factory = new WebStoreFactory();
+            var dbName = NewDbName();
+
+            factory.Create(BuildSchema(dbName)).Then<bool>(delegate(WebStoreClient client)
+            {
+                var table = client.Table<string, CrudEntity>(TableName);
+
+                table.UpSert(NewEntity("a", "todo", 1)).Then<bool>(delegate(string k1)
+                {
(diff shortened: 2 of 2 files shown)


Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
```diff
@@ -556,7 +570,13 @@ private void CursorIterator(
                 cursor.Continue();
             };
```
gautam-achieveai, line 574: **[Gautam's review bot]** MEDIUM -- `cursor.Continue()` and `cursor.Advance(skip)` remain unprotected

The PR's goal is to prevent *any* exception from escaping the `OnSuccess` handler. The user-callback seams are now covered, but these two IDB cursor-navigation calls are not:

- **`cursor.Continue()`** (this line) -- can throw `TransactionInactiveError` (transaction timed out or externally aborted) or `InvalidStateError`.
- **`cursor.Advance(skip)`** (line 534, not in diff) -- can throw `TransactionInactiveError` or `TypeError`.

If either throws, the exception escapes the handler and the `Promise` hangs -- the exact bug pattern this PR fixes for user callbacks.

**Suggested fix:**
```csharp
// Line 534 (Advance):
if (firstLoop)
{
    try { cursor.Advance(skip); }
    catch (Exception ex) { aborted = true; reject(ex); return; }
    firstLoop = false;
    return;
}

// Line 570 (Continue):
try { cursor.Continue(); }
catch (Exception ex) { aborted = true; reject(ex); }
```

These are rare edge cases (transaction timeout, browser quirks), but for consistency with the PR's stated goal of "no exception escapes OnSuccess", they should be covered.

gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of !43 (WI-41: wrap cursor visitor callbacks in try/catch -> reject)

---

## Context
**User Story #41**: WebStore cursor: wrap visitor callbacks in try/catch -> reject. Before this fix, a throwing user callback escaped the IDB `OnSuccess` handler, leaving the `Promise` permanently unsettled.

## Review Summary

Well-executed, focused PR. The try/catch barrier is correctly placed at the `CursorIterator` level -- the single centralization point through which all cursor-based operations flow. The `aborted` flag correctly prevents double-rejection and continued processing after failure. The 4 regression tests cover all high-risk user-callback seams across 4 different public API entry points (ForEach, Query, Delete, Update).

### What's done well
- **Correct centralization**: wrapping at `CursorIterator` means any new `*Internal` consumer gets protection automatically
- **`aborted` flag**: prevents double-settle and stops cursor iteration after catch -- all paths correctly avoid `cursor.Continue()` after abort
- **Exception preservation**: original exception passed directly to `reject(ex)`, no lossy wrapping
- **Test quality**: distinct error messages per test (`boom-visit`, `boom-filter`, etc.), resolve-path guards (`assert.IsTrue(false, ...)`), and side-effect verification for mutation operations (Delete/Update tests verify records survive intact)
- **PR description**: excellent -- explains the "one barrier instead of three" rationale clearly

### Findings

| # | Severity | Description |
|---|----------|-------------|
| 1 | MEDIUM | `cursor.Continue()` and `cursor.Advance()` not wrapped -- same hang risk as unprotected callbacks |
| 2 | MEDIUM | `OnError` wrapper: missing `aborted = true` + `evt.PreventDefault()` on early return |
| 3 | LOW | `TestForEachVisitThrowsRejects` could assert visitor called exactly once to verify abort stops iteration |

See inline comments for details.

### Verdict: **APPROVE with suggestions**
The two MEDIUM findings are defensive hardening for edge cases (IDB cursor API exceptions, browser-quirk double-fire). They don't block the PR but are worth a follow-up pass.

---
*Review agents used: exception-handling-review, test-coverage-review, duplicate-code-detector*

Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
```diff
@@ -556,7 +570,13 @@ private void CursorIterator(
...
 
-            request.OnError += HandleError(reject);
+            var onError = HandleError(reject);
+            request.OnError += (req, evt) =>
+            {
+                if (aborted)
+                { return; }
+                onError(req, evt);
+            };
         }
```
gautam-achieveai, line 585: **[Gautam's review bot]** MEDIUM -- Two defensive gaps in the `OnError` wrapper

**1. Missing `aborted = true` before `reject`:**
Every `catch` block in the `OnSuccess` handler sets `aborted = true` *before* calling `reject`. But this `OnError` path calls `reject` (via `onError`) without setting `aborted`. If a subsequent `OnSuccess` somehow fires (browser quirk or polyfill), it won't be guarded.

**2. Missing `evt.PreventDefault()` on early return:**
When `aborted == true`, the handler returns without calling `evt.PreventDefault()`. Compare with `HandleError` (line 325) and `AddBatchInternal`'s `OnError` (line 451) -- both call `PreventDefault()` unconditionally. Skipping it here could trigger a spurious `window.onerror` for the aborted IDB transaction error.

**Suggested fix:**
```csharp
request.OnError += (req, evt) =>
{
    evt.PreventDefault();
    if (aborted)
    { return; }
    aborted = true;
    onError(req, evt);
};
```

This matches the pattern in `AddBatchInternal` (lines 449-463) where `PreventDefault()` is called before the aborted check.

Test/Framework/Sunlight.Framework.Data.Test/WebStoreCrudTests.cs
```diff
@@ -657,6 +657,246 @@ public static void TestUniqueIndexEnforcesConstraint_DefectThree(Assert assert)
...
+                var table = client.Table<string, CrudEntity>(TableName);
+
+                table.UpSert(NewEntity("a", "todo", 1)).Then<bool>(delegate(string k1)
+                {
+                    table.UpSert(NewEntity("b", "todo", 2)).Then<bool>(delegate(string k2)
+                    {
+                        table.ForEach(Query.All, delegate(CrudEntity row)
+                        {
+                            throw new Exception("boom-visit");
+                        }).Then<bool, object>(
```
gautam-achieveai, line 687: **[Gautam's review bot]** LOW -- Consider asserting visitor is called exactly once

This test verifies the promise rejects with the correct error, but doesn't verify that the `aborted` flag actually *stops* cursor iteration for the second record. If the abort guard (line 521-522 in `CursorIterator`) were accidentally removed, the visitor would throw again for record "b" and `reject` would be called a second time -- but this test wouldn't detect it.

A visit counter would catch that regression:
```csharp
int visitCount = 0;
table.ForEach(Query.All, delegate(CrudEntity row)
{
    visitCount++;
    throw new Exception("boom-visit");
}).Then<bool, object>(
    ...
    delegate(object err)
    {
        assert.Equal(visitCount, 1, "visitor called exactly once before abort");
        // ... existing assertions ...
    });
```

Nice-to-have, not blocking.

Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 2d74f787. Wrapped both `cursor.Advance(skip)` and `cursor.Continue()` in try/catch following the suggested pattern — any IDB navigation exception now routes through `reject(ex)` with `aborted=true` instead of escaping `OnSuccess`.

Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 2d74f787. The `OnError` wrapper now mirrors the `AddBatchInternal` pattern: `evt.PreventDefault()` unconditionally, then `aborted` check, then `aborted = true` before dispatching to `onError`. Guards against browser-quirk `OnSuccess`-after-`OnError` and the spurious `window.onerror` scenario.

Test/Framework/Sunlight.Framework.Data.Test/WebStoreCrudTests.cs
gautam-achieveai: [Gautam's Dev bot]

Added in 2d74f787. `TestForEachVisitThrowsRejects` now asserts `visitCount == 1` on the reject path, locking in the behaviour that the `aborted` flag stops cursor iteration after the first throw. Regression on the abort guard would now fail this test.

gautam-achieveai (commented):
**[Gautam's review bot]** Re-review of !43, iteration 2 (commit 2d74f78)

---

## Context
**User Story #41**: WebStore cursor — wrap visitor callbacks in try/catch -> reject. Prevents throwing user callbacks from escaping the IDB `OnSuccess` handler and hanging the `Promise`.

## Previous Findings — All Verified

All 3 findings from iteration 1 have been addressed correctly in 2d74f787:

| # | Severity | Finding | Status | Verification |
|---|----------|---------|--------|-------------|
| 1 | MEDIUM | `cursor.Continue()` / `cursor.Advance()` unprotected | Fixed | Both wrapped in try/catch. `Advance` catch correctly includes `return` to skip `firstLoop = false`. `Continue` catch at end of handler — no `return` needed. |
| 2 | MEDIUM | `OnError` missing `PreventDefault()` + `aborted = true` | Fixed | Wrapper now mirrors `AddBatchInternal` pattern: `PreventDefault()` unconditional, aborted guard, `aborted = true`, dispatch. |
| 3 | LOW | `TestForEachVisitThrowsRejects` missing visit-count assertion | Fixed | `visitCount` variable added, asserts `== 1` on reject path — locks in abort-stops-iteration invariant. |

## New Issues in Iteration 2

**None.** The incremental diff is clean and minimal — exactly the fixes requested, nothing extraneous.

## Minor Observation (not blocking)

`evt.PreventDefault()` is called twice in the non-aborted `OnError` path — once in the wrapper (L578) and once inside `HandleError` (L325). This is **functionally harmless** (idempotent DOM call). The wrapper needs its own call for the `aborted` early-return path. Could be cleaned up by inlining reject (like `AddBatchInternal`), but not worth a round-trip.

## Verdict: **APPROVE**

All findings addressed. The `CursorIterator` now has comprehensive exception coverage: every user-callback seam, every IDB cursor navigation call, and the `OnError` handler all route exceptions through `reject()` with proper `aborted` flag coordination. The test suite validates the 4 highest-risk callback seams with regression coverage.

---
*Re-review agent: code-reviewer:exception-handling-review*

gautam-msft:
Not sure if it's OK to checkin known bug in the code???

Let's fix this:
---
Minor Observation (not blocking)
evt.PreventDefault() is called twice in the non-aborted OnError path — once in the wrapper (L578) and once inside HandleError (L325). This is functionally harmless (idempotent DOM call). The wrapper needs its own call for the aborted early-return path. Could be cleaned up by inlining reject (like AddBatchInternal), but not worth a round-trip.

gautam-achieveai:
[Gautam's Dev bot]

Cleaned up in c8de775f. Inlined `reject(new EventBasedException(...))` directly in the `OnError` wrapper (mirroring `AddBatchInternal`), so there's now a single `evt.PreventDefault()` call that covers both the aborted early-return path and the normal reject path. The indirection through `HandleError(reject)` is gone.

Data suite 26/26 tests still pass.

gautam-achieveai (commented):
**[Gautam's review bot]** Re-review of !43, iteration 3 (commit c8de775)

---

## Context
**User Story #41**: WebStore cursor — wrap visitor callbacks in try/catch → reject.

## Trigger
@gautam-msft requested cleanup of the double `PreventDefault()` in the non-aborted `OnError` path (iteration 2 observation).

## Iteration 3 Change — Verified ✅

The `HandleError(reject)` indirection has been replaced with an inlined `reject(new EventBasedException(...))`, exactly mirroring the `AddBatchInternal` pattern (L449-463):

```csharp
request.OnError += (req, evt) =>
{
    evt.PreventDefault();      // single call — covers both paths
    if (aborted) { return; }
    aborted = true;
    reject(new EventBasedException(
        req.Error.ExceptionMessage(),
        evt));
};
```

- **Single `PreventDefault()`** — no more double call. ✅
- **Reject arguments identical** to what `HandleError` produced. ✅
- **`HandleError` not dead code** — still used by 8 other call sites (Add, Count, Get, Delete, UpSert, etc.). ✅

## All Findings Across All Iterations

| # | Severity | Finding | Iteration Fixed | Status |
|---|----------|---------|-----------------|--------|
| 1 | MEDIUM | `cursor.Continue()` / `cursor.Advance()` unprotected | 2 (2d74f78) | ✅ |
| 2 | MEDIUM | `OnError` missing `PreventDefault()` + `aborted = true` | 2 (2d74f78) | ✅ |
| 3 | LOW | `visitCount` assertion missing in ForEach test | 2 (2d74f78) | ✅ |
| 4 | — | Double `PreventDefault()` (gautam-msft) | 3 (c8de775) | ✅ |

## Verdict: **APPROVE**

All findings from all 3 iterations addressed. The `CursorIterator` now has comprehensive exception coverage with clean, consistent patterns matching `AddBatchInternal`. No new issues.

---
*Review: manual inspection (7-line incremental diff)*

gautam-achieveai:
**[Gautam's review bot]** Verified the fix in c8de775 — the double `PreventDefault()` is gone. The `OnError` wrapper now inlines `reject(new EventBasedException(...))` directly, matching the `AddBatchInternal` pattern exactly. Single `PreventDefault()` call covers both the aborted early-return and normal reject paths. `HandleError` remains in use by 8 other call sites — no dead code.

All findings across all 3 iterations are now addressed. LGTM.

PR #44, opened 2026-04-25
[Gautam's Dev bot] WI-19: Repo-linked source maps (GitHub/ADO)

### LINKED ISSUE #19: Support linking source maps to ADO/GitHub repository files

gautam-achieveai:
Closes #19

## Summary

Wires NScript-generated source maps to host raw source content from a remote Git
repository (GitHub raw or Azure DevOps Items API) instead of (or in addition to)
the local `SourceMapFileHandler` / legacy `SrcMapper.ashx`. Browser DevTools
opening a `.map` whose `sourceRoot` is a public-repo raw URL gets the original
C# source from the SHA the JS was compiled from — no server-side source files
needed.

## What's in this PR

**Compiler (Phase A):**
- `RepoUrlBuilder` (`Sources/Compiler/SourceMap`) — provider detection +
  raw-URL construction for GitHub, ADO modern, ADO legacy `*.visualstudio.com`,
  and SSH remote variants.
- `SourceMap.RepoRoot` — when set, `sources[i]` are emitted as forward-slash,
  repo-relative paths so they combine cleanly with the remote `sourceRoot`.
  Files outside the repo root keep the legacy absolutized form.
- New `-repoRoot` CLI option, paired with `-sourceMapRoot` (rejected alone —
  partial config would silently produce broken maps).

**Build (Phase B):**
- `ComputeNScriptRepoMetadata` MSBuild target gated by `<RepoLinkedSourceMaps>true</RepoLinkedSourceMaps>`.
  Auto-resolves origin URL, commit SHA, and worktree root from `git`; explicit
  property overrides let CI feed `$(GITHUB_SHA)` / `$(BUILD_SOURCEVERSION)`
  directly without invoking `git` inside the build container.
- Inline `Regex.Match` property functions construct the same URL shapes as
  `RepoUrlBuilder` (kept under unit-test for drift).
- Origin URLs are redacted (`://[^/@]+@` → `://***@`) before being echoed into
  build logs so embedded PATs/passwords don't leak.

**Server (Phase C):**
- `SourceMapFileHandlerOptions.RepoUrlRedirectOnMiss` — when true, the handler
  issues `302 Found` to `{sourceRoot}{shortName}` instead of `404` whenever the
  local lookup misses. Gated:
  - `sourceRoot` MUST be `http://` or `https://` (legacy ashx, `javascript:`,
    `data:`, relative paths refused).
  - Short name MUST exist in the parsed map's `sources[]` array (refuses to
    redirect for paths the attacker dangles after the prefix).

**Tests (Phase D):**
- `RepoUrlBuilderTests` — 22 cases (GitHub/ADO modern/ADO legacy/SSH variants).
- `SourceMapTests` — repo-root rebase scenarios (in-repo, out-of-repo, mixed,
  altsep handling).
- `SourceMapFileHandlerTests` — redirect on/off, http(s) gate, sources[]
  membership gate, `ServeFromSourcesLong=false` interaction.
- `CompilerLogTests` — `-repoRoot` parsing (paired/unpaired/duplicated/missing-value).

**Docs (Phase E):**
- `docs/source-maps/repo-linked-source-maps.md` — end-to-end flow, CI setup,
  ADR-aligned design, threat model (tampered maps, credential leaks).

**E2E (Phase F):**
- `Test/Compiler/SourceMap.Server.E2E.Test` — Playwright MSTest suite gated by
  `ENABLE_PLAYWRIGHT_E2E` env var. Two-host Kestrel topology with ephemeral
  port binding via `IServerAddressesFeature`. Excluded from default test runs
  to avoid CI flakiness from browser downloads.

## Test plan

- [x] `dotnet build NScript_Full.sln -c Release` — 0 errors, 45 unrelated warnings
- [x] `SourceMap.Test` — 75/75 passing
- [x] `SourceMap.Server.Test` — 36/36 passing
- [x] `NScript.Utils.Test` — 28/28 passing
- [ ] Reviewer to sanity-check the `Sdk.targets` Regex property-function chain on a real consumer project (requires the new `Mcqdb.NScript.Sdk` package version)
- [ ] Optional: enable `ENABLE_PLAYWRIGHT_E2E=1` and run the E2E suite locally

## Out of scope

- Authenticated source-fetching for private repos (the redirect hands off to the
  browser's session cookies; no proxying).
- Indexing / caching of remote source content (always fresh fetch).
- Alternative providers (GitLab, Bitbucket) — `RepoUrlBuilder.Provider` is
  open-ended for future extension.

### DIFF
--- a/Sources/Compiler/NScript.Sdk/Sdk/Sdk.targets
+++ b/Sources/Compiler/NScript.Sdk/Sdk/Sdk.targets
@@ -10,20 +10,112 @@
 		<Minify Condition="$(Minify) == ''">false</Minify>
 		<Uglify Condition="$(Uglify) == ''">false</Uglify>
 		<JsOptimize Condition="$(JsOptimize) == ''">false</JsOptimize>
+		<RepoLinkedSourceMaps Condition="'$(RepoLinkedSourceMaps)' == ''">false</RepoLinkedSourceMaps>
 	</PropertyGroup>
 
-	<Target Name="AfterCompile" DependsOnTargets="$(CoreCompile)" > 
+	<Target Name="AfterCompile" DependsOnTargets="$(CoreCompile)" >
 		<MsBuild
 			Condition="$(GenerateJs) == True"
 			Projects="$(MSBuildProjectFile)"
-			Targets="ScriptGenerate" /> 
+			Targets="ScriptGenerate" />
+	</Target>
+
+	<!--
+	  Captures the Git remote URL, commit SHA, and worktree root, then composes a raw-file
+	  base URL for either GitHub or Azure DevOps. Runs only when <RepoLinkedSourceMaps>true</RepoLinkedSourceMaps>
+	  is set on the project. Each git invocation tolerates failure (ContinueOnError) so a build
+	  outside a working tree (CI cache restore, source-only ZIP) emits a warning and skips the
+	  rest of the target rather than failing.
+
+	  Auto-detection prefers explicit overrides (<SourceRepoOriginUrl>, <SourceRepoCommitSha>,
+	  <NScriptRepoRoot>) so CI can feed values directly from $(GITHUB_SHA) /
+	  $(BUILD_SOURCEVERSION) without invoking git inside the build container.
+
+	  URL construction uses MSBuild property functions (Regex.Match) so we don't ship a
+	  separate task assembly; the same regex patterns are kept under unit test in
+	  RepoUrlBuilder.cs (Sources/Compiler/SourceMap).
+	  -->
+	<Target Name="ComputeNScriptRepoMetadata"
+	        Condition="'$(RepoLinkedSourceMaps)' == 'true' and '$(SourceMapRoot)' == ''"
+	        BeforeTargets="ScriptGenerate">
+
+		<Exec Command="git remote get-url origin"
+		      Condition="'$(SourceRepoOriginUrl)' == ''"
+		      ConsoleToMSBuild="true"
+		      EchoOff="true"
+		      IgnoreExitCode="true"
+		      ContinueOnError="WarnAndContinue">
+			<Output TaskParameter="ConsoleOutput" PropertyName="_NScriptGitRemoteUrlRaw" />
+			<Output TaskParameter="ExitCode" PropertyName="_NScriptGitRemoteExit" />
+		</Exec>
+
+		<Exec Command="git rev-parse HEAD"
+		      Condition="'$(SourceRepoCommitSha)' == ''"
+		      ConsoleToMSBuild="true"
+		      EchoOff="true"
+		      IgnoreExitCode="true"
+		      ContinueOnError="WarnAndContinue">
+			<Output TaskParameter="ConsoleOutput" PropertyName="_NScriptGitShaRaw" />
+			<Output TaskParameter="ExitCode" PropertyName="_NScriptGitShaExit" />
+		</Exec>
+
+		<Exec Command="git rev-parse --show-toplevel"
+		      Condition="'$(NScriptRepoRoot)' == ''"
+		      ConsoleToMSBuild="true"
+		      EchoOff="true"
+		      IgnoreExitCode="true"
+		      ContinueOnError="WarnAndContinue">
+			<Output TaskParameter="ConsoleOutput" PropertyName="_NScriptGitTopLevelRaw" />
+		</Exec>
+
+		<PropertyGroup>
+			<SourceRepoOriginUrl Condition="'$(SourceRepoOriginUrl)' == '' and '$(_NScriptGitRemoteUrlRaw)' != ''">$(_NScriptGitRemoteUrlRaw.Trim())</SourceRepoOriginUrl>
+			<SourceRepoCommitSha Condition="'$(SourceRepoCommitSha)' == '' and '$(_NScriptGitShaRaw)' != ''">$(_NScriptGitShaRaw.Trim())</SourceRepoCommitSha>
+			<NScriptRepoRoot Condition="'$(NScriptRepoRoot)' == '' and '$(_NScriptGitTopLevelRaw)' != ''">$(_NScriptGitTopLevelRaw.Trim())</NScriptRepoRoot>
+			<SourceRepoProvider Condition="'$(SourceRepoProvider)' == ''">auto</SourceRepoProvider>
+
+			<!--
+			  Some developers / CI systems bake credentials into the origin URL
+			  (https://token@github.com/..., https://user:pass@dev.azure.com/...).
+			  Strip everything between '://' and the first '@' before echoing the URL into
+			  build logs so a stray PAT or password doesn't end up in CI artifacts.
+			  The unredacted value is still used by the regex matchers above (which only
+			  read host/path groups).
+			-->
+			<_NScriptOriginUrlRedacted Condition="'$(SourceRepoOriginUrl)' != ''">$([System.Text.RegularExpressions.Regex]::Replace('$(SourceRepoOriginUrl)', '://[^/@]+@', '://***@'))</_NScriptOriginUrlRedacted>
+		</PropertyGroup>
+
+		<Warning Text="RepoLinkedSourceMaps requested but git metadata could not be captured (origin='$(_NScriptOriginUrlRedacted)' sha='$(SourceRepoCommitSha)'). Skipping repo-linked sourceRoot."
+		         Condition="'$(SourceRepoOriginUrl)' == '' or '$(SourceRepoCommitSha)' == ''" />
+
+		<!-- GitHub detection: matches https://github.com/{owner}/{repo}[.git] AND git@github.com:{owner}/{repo}[.git] -->
+		<PropertyGroup Condition="'$(SourceRepoOriginUrl)' != '' and '$(SourceRepoCommitSha)' != '' and ('$(SourceRepoProvider)' == 'auto' or '$(SourceRepoProvider)' == 'GitHub')">
+			<_NScriptGitHubOwner>$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '(?i)github\.com[/:]([^/]+)/([^/]+?)(?:\.git)?/?$').Groups[1].Value)</_NScriptGitHubOwner>
+			<_NScriptGitHubRepo>$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '(?i)github\.com[/:]([^/]+)/([^/]+?)(?:\.git)?/?$').Groups[2].Value)</_NScriptGitHubRepo>
+			<SourceMapRoot Condition="'$(_NScriptGitHubOwner)' != '' and '$(_NScriptGitHubRepo)' != ''">https://raw.githubusercontent.com/$(_NScriptGitHubOwner)/$(_NScriptGitHubRepo)/$(SourceRepoCommitSha)/</SourceMapRoot>
+		</PropertyGroup>
+
+		<!-- Azure DevOps modern: https://dev.azure.com/{org}/{project}/_git/{repo} -->
+		<PropertyGroup Condition="'$(SourceMapRoot)' == '' and '$(SourceRepoOriginUrl)' != '' and '$(SourceRepoCommitSha)' != '' and ('$(SourceRepoProvider)' == 'auto' or '$(SourceRepoProvider)' == 'AzureDevOps')">
+			<_NScriptAdoOrg>$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '(?i)dev\.azure\.com/([^/]+)/([^/]+)/_git/([^/]+?)(?:\.git)?/?$').Groups[1].Value)</_NScriptAdoOrg>
+			<_NScriptAdoProject>$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '(?i)dev\.azure\.com/([^/]+)/([^/]+)/_git/([^/]+?)(?:\.git)?/?$').Groups[2].Value)</_NScriptAdoProject>
+			<_NScriptAdoRepo>$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '(?i)dev\.azure\.com/([^/]+)/([^/]+)/_git/([^/]+?)(?:\.git)?/?$').Groups[3].Value)</_NScriptAdoRepo>
+			<SourceMapRoot Condition="'$(_NScriptAdoOrg)' != '' and '$(_NScriptAdoProject)' != '' and '$(_NScriptAdoRepo)' != ''">https://dev.azure.com/$(_NScriptAdoOrg)/$(_NScriptAdoProject)/_apis/git/repositories/$(_NScriptAdoRepo)/items?api-version=7.1&amp;versionDescriptor.version=$(SourceRepoCommitSha)&amp;versionDescriptor.versionType=commit&amp;path=/</SourceMapRoot>
+		</PropertyGroup>
+
+		<Warning Text="RepoLinkedSourceMaps requested but origin URL '$(_NScriptOriginUrlRedacted)' did not match GitHub or Azure DevOps; emitting map without repo-linked sourceRoot."
+		         Condition="'$(SourceMapRoot)' == '' and '$(SourceRepoOriginUrl)' != '' and '$(SourceRepoCommitSha)' != ''" />
+
+		<Message Importance="high"
+		         Text="NScript repo-linked source maps: sourceRoot='$(SourceMapRoot)' repoRoot='$(NScriptRepoRoot)'"
+		         Condition="'$(SourceMapRoot)' != ''" />
 	</Target>
 
 	<Target
 		Name="ScriptGenerate"
 		Inputs="$(BaseIntermediateOutputPath)\$(Configuration)\$(TargetFramework)\$(AssemblyName).$(Ext);@(ReferencePath)"
 		Outputs="$(JsOutputPath)\$(AssemblyName).js"
-		DependsOnTargets="$(CoreCompile)"
+		DependsOnTargets="$(CoreCompile);ComputeNScriptRepoMetadata"
 		Condition="$(GenerateJs) == True">
 
 		<Message
@@ -42,6 +134,7 @@
 			<ScriptGenerateOption Include="-uglify" Condition="'$(Uglify)' == 'true'" />
 			<ScriptGenerateOption Include="-optimize" Condition="'$(JsOptimize)' == 'true'" />
 			<ScriptGenerateOption Include="-sourceMapRoot &quot;$(SourceMapRoot)&quot;" Condition="'$(SourceMapRoot)' != ''" />
+			<ScriptGenerateOption Include="-repoRoot &quot;$(NScriptRepoRoot)&quot;" Condition="'$(SourceMapRoot)' != '' and '$(NScriptRepoRoot)' != ''" />
 		</ItemGroup>
 
 		<MakeDir Directories="$(JsOutputPath)" Condition="!Exists('$(JsOutputPath)')" />
--- a/Sources/Compiler/NScript.Sdk/Sdk/Sdk.targets
+++ b/Sources/Compiler/NScript.Sdk/Sdk/Sdk.targets
@@ -88,18 +88,51 @@
 		<Warning Text="RepoLinkedSourceMaps requested but git metadata could not be captured (origin='$(_NScriptOriginUrlRedacted)' sha='$(SourceRepoCommitSha)'). Skipping repo-linked sourceRoot."
 		         Condition="'$(SourceRepoOriginUrl)' == '' or '$(SourceRepoCommitSha)' == ''" />
 
-		<!-- GitHub detection: matches https://github.com/{owner}/{repo}[.git] AND git@github.com:{owner}/{repo}[.git] -->
+		<!--
+		  GitHub HTTPS: https://[creds@]github.com/{owner}/{repo}[.git]
+		  GitHub SSH (modern):   git@github.com:{owner}/{repo}[.git]
+		  GitHub SSH (ssh://):   ssh://git@github.com/{owner}/{repo}[.git]
+		  Both SSH forms collapse onto a single regex by allowing `[/:]` after `github.com`
+		  and tolerating the optional `ssh://git@` / `git@` prefix anywhere before the host.
+		  Keep these patterns in sync with `RepoUrlBuilder.cs` (covered by RepoUrlBuilderTests).
+		-->
 		<PropertyGroup Condition="'$(SourceRepoOriginUrl)' != '' and '$(SourceRepoCommitSha)' != '' and ('$(SourceRepoProvider)' == 'auto' or '$(SourceRepoProvider)' == 'GitHub')">
-			<_NScriptGitHubOwner>$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '(?i)github\.com[/:]([^/]+)/([^/]+?)(?:\.git)?/?$').Groups[1].Value)</_NScriptGitHubOwner>
-			<_NScriptGitHubRepo>$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '(?i)github\.com[/:]([^/]+)/([^/]+?)(?:\.git)?/?$').Groups[2].Value)</_NScriptGitHubRepo>
+			<_NScriptGitHubHttpsPattern>(?i)^https?://(?:[^/@]+@)?github\.com/([^/]+)/([^/]+?)(?:\.git)?/?$</_NScriptGitHubHttpsPattern>
+			<_NScriptGitHubSshPattern>(?i)^(?:ssh://)?git@github\.com[:/]([^/]+)/([^/]+?)(?:\.git)?/?$</_NScriptGitHubSshPattern>
+
+			<_NScriptGitHubOwner>$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '$(_NScriptGitHubHttpsPattern)').Groups[1].Value)</_NScriptGitHubOwner>
+			<_NScriptGitHubRepo>$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '$(_NScriptGitHubHttpsPattern)').Groups[2].Value)</_NScriptGitHubRepo>
+
+			<_NScriptGitHubOwner Condition="'$(_NScriptGitHubOwner)' == ''">$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '$(_NScriptGitHubSshPattern)').Groups[1].Value)</_NScriptGitHubOwner>
+			<_NScriptGitHubRepo Condition="'$(_NScriptGitHubRepo)' == ''">$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '$(_NScriptGitHubSshPattern)').Groups[2].Value)</_NScriptGitHubRepo>
+
 			<SourceMapRoot Condition="'$(_NScriptGitHubOwner)' != '' and '$(_NScriptGitHubRepo)' != ''">https://raw.githubusercontent.com/$(_NScriptGitHubOwner)/$(_NScriptGitHubRepo)/$(SourceRepoCommitSha)/</SourceMapRoot>
 		</PropertyGroup>
 
-		<!-- Azure DevOps modern: https://dev.azure.com/{org}/{project}/_git/{repo} -->
+		<!--
+		  Azure DevOps modern HTTPS: https://[creds@]dev.azure.com/{org}/{project}/_git/{repo}
+		  Azure DevOps modern SSH:   git@ssh.dev.azure.com:v3/{org}/{project}/{repo}
+		  Azure DevOps legacy HTTPS: https://[creds@]{org}.visualstudio.com/{project}/_git/{repo}
+		  All three flow into the same Items API URL shape (org / project / repo / sha / path=/).
+		  Keep these patterns in sync with `RepoUrlBuilder.cs` (covered by RepoUrlBuilderTests).
+		-->
 		<PropertyGroup Condition="'$(SourceMapRoot)' == '' and '$(SourceRepoOriginUrl)' != '' and '$(SourceRepoCommitSha)' != '' and ('$(SourceRepoProvider)' == 'auto' or '$(SourceRepoProvider)' == 'AzureDevOps')">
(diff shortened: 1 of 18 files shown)


gautam-achieveai (commented):
## [Gautam's review bot] Deep Review — PR #44: Repo-Linked Source Maps (WI-19)

### Context
**User Story #19**: Support linking source maps to ADO/GitHub repository files.
Feature is opt-in via `<RepoLinkedSourceMaps>true</RepoLinkedSourceMaps>`, constructs `sourceRoot` URLs pointing to `raw.githubusercontent.com` (GitHub) or ADO Items API, with optional 302 redirect fallback in the ASP.NET Core handler.

### Overall Assessment
Well-structured feature PR with good security posture, thorough documentation, and solid test coverage across 17 files (~1500 insertions). The architecture respects layer boundaries (compiler vs server), the opt-in design prevents impact on existing builds, and the threat model documentation is thoughtful. The commit message is exemplary.

However, the review identified **4 HIGH**, **4 MEDIUM**, and **4 LOW** findings — three of which are blockers that should be addressed before merge. All three blocker fixes are straightforward.

### Review Agents Used
Architecture · Security · Performance · Test Coverage · Exception Handling · Review Grader

---

### :red_circle: HIGH — Blockers (3)

**H1. Sdk.targets regex only handles `dev.azure.com` HTTPS — misses ADO legacy + SSH + GitHub ssh:// prefix**
`RepoUrlBuilder.cs` handles 5 URL variants (GitHub HTTPS/SSH, ADO modern HTTPS/SSH, ADO legacy HTTPS). `Sdk.targets` only handles 2 (GitHub `[/:]` and ADO modern HTTPS `dev.azure.com`). Engineers using `*.visualstudio.com` or `git@ssh.dev.azure.com:v3/` get a confusing build warning and no `sourceRoot`. The code comment says "keep in sync" — they aren't. See inline comment.

**H2. Missing `SecurityException` catches at 4 `Path.GetFullPath`/`FileInfo` sites in the handler**
This PR establishes a 4-exception pattern (`ArgumentException` + `PathTooLongException` + `NotSupportedException` + `SecurityException`) at `NormalizeRepoRoot`, `TryRebaseToRepoRoot`, `ResolveMapsDirectory`, `ResolveAllowedRoots`. But 4 other catch sites in `SourceMapFileHandler.cs` and `SourceMapSources.cs` omit `SecurityException`. In CAS-restricted environments, this surfaces as 500 instead of 404. See inline comment.

**H3. `TryBuildRepoRedirect` is a security boundary method with zero direct tests**
Marked `internal` specifically for testing, but never tested directly. The hostile-scheme rejection (`javascript:`, `data:`, `file://`) — the security-critical branch — has no assertions. Add `[DataRow]` tests for hostile schemes + `null sources` + empty `sourceRoot`.

### :yellow_circle: MEDIUM — Should Fix (4)

**M1. Unnecessary map parsing when both `ServeFromSourcesLong=false` AND `RepoUrlRedirectOnMiss=false`**
Map is always parsed before checking `ServeFromSourcesLong`. When neither consumer is enabled (the default), every request wastes file I/O + JSON parse (up to 16MB). Fix: `bool needsParse = options.ServeFromSourcesLong || options.RepoUrlRedirectOnMiss;` See inline comment.

**M2. Unprotected `Path.GetFullPath` fallback in `SourceMap.ToString()`**
`TryRebaseToRepoRoot` safely catches exceptions and returns null. The null-coalescing fallback calls `Path.GetFullPath` without try-catch. Pre-existing pattern but the PR introduces a new contract where null = "use fallback" — the fallback should be equally safe. See inline comment.

**M3. `OrdinalIgnoreCase` path containment on case-sensitive filesystems**
`TryRebaseToRepoRoot` uses `StringComparison.OrdinalIgnoreCase` while the handler's `IsContained` correctly uses platform-specific comparison. On Linux, `/Repo/foo` could match repo root `/repo/` when they're different directories. See inline comment.

**M4. Missing direct unit tests for `NormalizeRepoRoot`, `TryRebaseToRepoRoot`, `SourceMapSources.SourceRoot`**
These internal methods have non-trivial exception handling (4 catch blocks each) and escape/unescape roundtrip logic. Only tested indirectly through `ToString()`. `SourceRoot` parsing has zero test assertions despite being the data foundation for the redirect feature.

### :blue_circle: LOW — Consider (4)

**L1.** `http://` allowed in redirect — no legitimate use case for plain HTTP targets. Consider restricting to `https://` only.
**L2.** Missing URL encoding on `redirectUrl = root + sourceName` — a `sourceName` containing `#` could truncate ADO's `path=/` query param at the fragment boundary.
**L3.** E2E test project not included in `NScript_Full.sln` — likely intentional but worth a comment in the csproj.
**L4.** `_NScriptGitRemoteExit` property captured but never checked in Sdk.targets (dead code).

- Clean layer separation: `RepoUrlBuilder` in compiler-side `SourceMap`, redirect in `SourceMap.Server`, no cross-dependency
- MSBuild target gracefully degrades with `ContinueOnError="WarnAndContinue"` + credential redaction
- `RepoUrlRedirectOnMiss` is opt-in + protocol-restricted + sources[]-membership-gated
- Thorough test coverage: 25+ new tests across 5 files
- Excellent documentation with threat model, CI config, and limitations section
- `sourcesLong` preserved for local-disk resolution alongside the new repo-relative `sources[]`

### Suggested Refactoring (Non-blocking)
- Extract `SourceMapOptions` record to replace the 7-parameter `JSWriter.Write` private method. Threading `repoRoot` as a raw string through 4 layers (ParseOptions → Builder → JSWriter → SourceMap) will get worse with each new source-map option.

Sources/Compiler/NScript.Sdk/Sdk/Sdk.targets
```diff
@@ -10,20 +10,112 @@
...
+		         Condition="'$(SourceRepoOriginUrl)' == '' or '$(SourceRepoCommitSha)' == ''" />
+
+		<!-- GitHub detection: matches https://github.com/{owner}/{repo}[.git] AND git@github.com:{owner}/{repo}[.git] -->
+		<PropertyGroup Condition="'$(SourceRepoOriginUrl)' != '' and '$(SourceRepoCommitSha)' != '' and ('$(SourceRepoProvider)' == 'auto' or '$(SourceRepoProvider)' == 'GitHub')">
+			<_NScriptGitHubOwner>$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '(?i)github\.com[/:]([^/]+)/([^/]+?)(?:\.git)?/?$').Groups[1].Value)</_NScriptGitHubOwner>
+			<_NScriptGitHubRepo>$([System.Text.RegularExpressions.Regex]::Match('$(SourceRepoOriginUrl)', '(?i)github\.com[/:]([^/]+)/([^/]+?)(?:\.git)?/?$').Groups[2].Value)</_NScriptGitHubRepo>
+			<SourceMapRoot Condition="'$(_NScriptGitHubOwner)' != '' and '$(_NScriptGitHubRepo)' != ''">https://raw.githubusercontent.com/$(_NScriptGitHubOwner)/$(_NScriptGitHubRepo)/$(SourceRepoCommitSha)/</SourceMapRoot>
+		</PropertyGroup>
+
+		<!-- Azure DevOps modern: https://dev.azure.com/{org}/{project}/_git/{repo} -->
```
gautam-achieveai: **[Gautam's review bot] :red_circle: HIGH — Sdk.targets regex misses 3 of 5 URL variants**

`RepoUrlBuilder.cs` handles:
1. :white_check_mark: GitHub HTTPS (`github.com/owner/repo`)
2. :x: GitHub SSH prefix (`ssh://git@github.com/owner/repo.git`) — the `[/:]` trick covers `git@github.com:` but NOT `ssh://git@github.com/`
3. :white_check_mark: ADO modern HTTPS (`dev.azure.com/{org}/{project}/_git/{repo}`)
4. :x: ADO modern SSH (`git@ssh.dev.azure.com:v3/{org}/{project}/{repo}`)
5. :x: ADO legacy HTTPS (`{org}.visualstudio.com/{project}/_git/{repo}`)

Sdk.targets only has regex blocks for #1 and #3. Engineers using `*.visualstudio.com` (common in established orgs) or SSH remotes get the confusing warning:
```
RepoLinkedSourceMaps requested but origin URL '…' did not match GitHub or Azure DevOps
```

`RepoUrlBuilder.cs` line 17 says: *"keep the regex patterns here and in Sdk.targets in sync, and rely on RepoUrlBuilderTests to catch drift"* — but the tests only cover the C# class, not the MSBuild regex. The drift is already real.

**Fix:** Add PropertyGroup blocks for ADO legacy HTTPS and ADO SSH, and extend the GitHub regex to handle the `ssh://` prefix. Or better: use an inline MSBuild task (`RoslynCodeTaskFactory`) that calls `RepoUrlBuilder.TryBuildRawUrl` directly — single source of truth.

Sources/Compiler/SourceMap.Server/SourceMapFileHandler.cs
```diff
@@ -278,6 +281,83 @@ private static async Task HandleAsync(
...
+
+            string root = sources.SourceRoot;
+            if (string.IsNullOrEmpty(root))
+            {
+                return false;
+            }
+
+            // Refuse to redirect to anything other than http(s) — the legacy `{file}.ashx` root,
+            // a relative path, or a hostile `javascript:`/`data:` URI would all be unsafe.
+            if (!root.StartsWith("http://", StringComparison.OrdinalIgnoreCase)
```
gautam-achieveai: **[Gautam's review bot] :blue_circle: LOW — Consider restricting to `https://` only**

This accepts both `http://` and `https://` sourceRoot URLs. The feature is opt-in and the docs clearly acknowledge the security model, but there's no legitimate use case for plain HTTP redirect targets — both GitHub and ADO only serve over HTTPS. Allowing `http://` creates unnecessary MITM surface for source code content.

Non-blocking, but something to consider tightening.

Sources/Compiler/SourceMap/SourceMap.cs
```diff
@@ -193,6 +201,98 @@ public string MapFile
...
+            catch (NotSupportedException)
+            {
+                return null;
+            }
+            catch (System.Security.SecurityException)
+            {
+                return null;
+            }
+
+            if (!fullPath.StartsWith(normalizedRepoRoot, StringComparison.OrdinalIgnoreCase))
```
gautam-achieveai: **[Gautam's review bot] :yellow_circle: MEDIUM — `OrdinalIgnoreCase` on case-sensitive filesystems**

This uses `StringComparison.OrdinalIgnoreCase` unconditionally. On Linux (case-sensitive), `/Repo/foo.cs` would match repo root `/repo/` even though they are different directories — producing incorrect repo-relative paths in the emitted map.

The handler's own `IsContained` (SourceMapFileHandler.cs:398-400) correctly uses platform-specific comparison:
```csharp
StringComparison cmp = OperatingSystem.IsWindows()
    ? StringComparison.OrdinalIgnoreCase
    : StringComparison.Ordinal;
```

**Fix:** Apply the same platform-specific comparison here.

Sources/Compiler/SourceMap.Server/SourceMapFileHandler.cs
```diff
@@ -161,17 +164,17 @@ private static async Task HandleAsync(
...
+                await NotFoundOrRedirectAsync(ctx, sources, sourceName, options);
                 return;
             }
             catch (PathTooLongException)
             {
-                ctx.Response.StatusCode = (int)HttpStatusCode.NotFound;
+                await NotFoundOrRedirectAsync(ctx, sources, sourceName, options);
                 return;
             }
             catch (NotSupportedException)
```
gautam-achieveai, line 179: **[Gautam's review bot] :red_circle: HIGH — Missing `SecurityException` catch**

This `Path.GetFullPath` catch block handles `ArgumentException`, `PathTooLongException`, `NotSupportedException` — but omits `SecurityException`. The same PR establishes the 4-exception convention at `NormalizeRepoRoot()` (SourceMap.cs:222-234), `TryRebaseToRepoRoot()` (SourceMap.cs:270-282), `ResolveMapsDirectory()`, and `ResolveAllowedRoots()`. This site and three others break the convention:

1. Here (line 165-178) — `Path.GetFullPath(longPath)`
2. Lines 196-210 — `new FileInfo(resolvedLongPath)` 
3. `SourceMapSources.TryParseAsync` lines 98-113 — `new FileInfo(mapFilePath)`
4. `TryGetFileSize` lines 361-387 — also missing `ArgumentException`/`NotSupportedException`

In CAS-restricted or containerized hosting, a `SecurityException` here surfaces as an unhandled 500 instead of the intended 404/redirect.

**Fix:** Add `catch (SecurityException)` with the same `NotFoundOrRedirectAsync` handler at all four sites.

Sources/Compiler/SourceMap.Server/SourceMapFileHandler.cs
```diff
@@ -128,19 +129,21 @@ private static async Task HandleAsync(
                 return;
             }
 
-            if (!options.ServeFromSourcesLong)
+            // Parse the map up-front. Both the local-file branch and the repo-redirect fallback
+            // need its sources/sourceRoot data, so a single parse keeps them aligned and avoids
+            // repeated disk I/O.
+            SourceMapSources sources = null;
+            if (TryGetFileSize(mapPath, out long mapSize) && mapSize <= options.MaxMapFileSizeBytes)
```
gautam-achieveai: **[Gautam's review bot] :yellow_circle: MEDIUM — Unnecessary map parsing when both features are off**

The map is now always parsed up-front (before checking `ServeFromSourcesLong`). When `ServeFromSourcesLong=false` AND `RepoUrlRedirectOnMiss=false` (both defaults), the parsed result is immediately discarded at line 143. For a typical 500KB map, each request wastes ~1MB UTF-16 string + JSON DOM + dictionary allocation.

This is a DevTools handler (low concurrency), so not a production outage risk. But 100% of requests in the `ServeFromSourcesLong=false` deployment mode hit this wasted path, and maps can be up to the 16MB cap.

**Fix:**
```csharp
bool needsParse = options.ServeFromSourcesLong || options.RepoUrlRedirectOnMiss;
SourceMapSources sources = null;
if (needsParse && TryGetFileSize(mapPath, out long mapSize) && mapSize <= options.MaxMapFileSizeBytes)
{
    sources = await SourceMapSources.TryParseAsync(...);
}
```

Sources/Compiler/SourceMap/SourceMap.cs
```diff
@@ -292,10 +392,12 @@ public override string ToString()
             if (this.files.Count > 0)
             {
                 Dictionary<string, int> fileMap = new Dictionary<string, int>(this.files.Count);
+                string normalizedRepoRoot = NormalizeRepoRoot(this.RepoRoot);
                 sb.Append("\t\"sources\": [");
                 for (int i = 0; i < this.files.Count; i++)
                 {
-                    var fileName = Path.GetFullPath(this.files[i]).Replace(":", "$").Replace("\\", "/");
+                    var fileName = TryRebaseToRepoRoot(this.files[i], normalizedRepoRoot)
+                        ?? Path.GetFullPath(this.files[i]).Replace(":", "$").Replace("\\", "/");
```
gautam-achieveai: **[Gautam's review bot] :yellow_circle: MEDIUM — Unprotected `Path.GetFullPath` fallback**

`TryRebaseToRepoRoot` correctly catches `ArgumentException`, `PathTooLongException`, `NotSupportedException`, `SecurityException` and returns null. But the null-coalescing fallback here calls `Path.GetFullPath(this.files[i])` WITHOUT any try-catch. If the path is malformed, this throws and crashes the map-writing step.

The unprotected call is pre-existing, but the PR introduces a new contract where `TryRebaseToRepoRoot` returning null means "use the fallback" — including when it returned null because `GetFullPath` threw internally. In that case, a slightly different input (`this.files[i]` with escaped backslashes) is passed to the unprotected call — it will likely also throw.

**Fix:** Wrap the fallback in a try-catch, or extract a `TryGetFullPath` helper:
```csharp
string fileName = TryRebaseToRepoRoot(this.files[i], normalizedRepoRoot);
if (fileName == null)
{
    try { fileName = Path.GetFullPath(this.files[i]).Replace(":", "$").Replace("\\", "/"); }
    catch { fileName = this.files[i].Replace("\\\\", "/"); }
}
```

Sources/Compiler/NScript.Sdk/Sdk/Sdk.targets
gautam-achieveai: [Gautam's Dev bot]

Fixed in 1996150. Extended both PropertyGroup regex blocks in `Sdk.targets` to mirror all 5 patterns from `RepoUrlBuilder.cs`:

- GitHub: `_NScriptGitHubHttpsPattern` (https with optional creds) + `_NScriptGitHubSshPattern` (covers both `git@github.com:` and `ssh://git@github.com/` via `[:/]`).
- ADO: `_NScriptAdoModernHttpsPattern`, `_NScriptAdoModernSshPattern` (`git@ssh.dev.azure.com:v3/...`), and `_NScriptAdoLegacyHttpsPattern` (`{org}.visualstudio.com`).

Each provider tries patterns in order via `Condition="'$(_NScriptXxx)' == ''"`, no new task assembly. Disposition: Fixed.

Sources/Compiler/SourceMap.Server/SourceMapFileHandler.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 1996150. `TryBuildRepoRedirect` now refuses anything other than `https://` — both GitHub raw and the ADO Items API are HTTPS-only, and a tampered map shouldn't be able to silently downgrade DevTools to cleartext. Added a direct test (`TryBuildRepoRedirect_HttpSourceRoot_ReturnsFalse`). Disposition: Fixed.

Sources/Compiler/SourceMap/SourceMap.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 1996150. `TryRebaseToRepoRoot` now selects the comparer based on `OperatingSystem.IsWindows()` (matching the pattern used in `SourceMapFileHandler.IsContained`):

```csharp
StringComparison cmp = OperatingSystem.IsWindows()
    ? StringComparison.OrdinalIgnoreCase
    : StringComparison.Ordinal;
```

This keeps Windows behaviour unchanged while preventing case-collision false positives on Linux/macOS. Disposition: Fixed.

Sources/Compiler/SourceMap.Server/SourceMapFileHandler.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 1996150. Added `SecurityException` catches at the four sites that were missing them:

1. `SourceMapFileHandler.HandleAsync` — `Path.GetFullPath(longPath)` block.
2. `SourceMapFileHandler.HandleAsync` — `new FileInfo(resolvedLongPath)` block.
3. `SourceMapSources.TryParseAsync` — `new FileInfo(mapFilePath)` block.
4. `SourceMapFileHandler.TryGetFileSize` — also rounded out with `ArgumentException`, `PathTooLongException`, and `NotSupportedException` so the 4-exception parity holds across `NormalizeRepoRoot` / `TryRebaseToRepoRoot` / `ResolveMapsDirectory` / `ResolveAllowedRoots` and the `FileInfo` constructors used at request time.

Each catch follows the existing pattern at its site (calls `NotFoundOrRedirectAsync` or returns the local fallback). Disposition: Fixed.

Sources/Compiler/SourceMap.Server/SourceMapFileHandler.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 1996150:

```csharp
bool needsParse = options.ServeFromSourcesLong || options.RepoUrlRedirectOnMiss;
if (needsParse
    && TryGetFileSize(mapPath, out long mapSize)
    && mapSize <= options.MaxMapFileSizeBytes)
{
    sources = await SourceMapSources.TryParseAsync(mapPath, options.MaxMapFileSizeBytes, ctx.RequestAborted);
}
```

When both feature flags are off (the default-config request path), the parse is skipped entirely — no `FileInfo`, no `FileStream`, no JSON document — and the `if (!options.ServeFromSourcesLong)` branch below returns 404 immediately. Disposition: Fixed.

Sources/Compiler/SourceMap/SourceMap.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 1996150. The fallback is now extracted into `AbsolutizeForSources` and wrapped in the same 4-exception set used elsewhere (`ArgumentException`, `PathTooLongException`, `NotSupportedException`, `SecurityException`). On any of those failures it degrades to `rawFile.Replace("\\\\", "/")` (the unprocessed forward-slashed form) instead of letting the throw escape from `ToString()`.

```csharp
private static string AbsolutizeForSources(string rawFile)
{
    try { return Path.GetFullPath(rawFile).Replace(":", "$").Replace("\\", "/"); }
    catch (ArgumentException)            { return rawFile.Replace("\\\\", "/"); }
    catch (PathTooLongException)         { return rawFile.Replace("\\\\", "/"); }
    catch (NotSupportedException)        { return rawFile.Replace("\\\\", "/"); }
    catch (System.Security.SecurityException) { return rawFile.Replace("\\\\", "/"); }
}
```

Disposition: Fixed.

gautam-achieveai:
[Gautam's Dev bot]

Iter-1 review feedback addressed in 19961501. Build green, SourceMap.Test 85/85 pass, SourceMap.Server.Test 60/60 pass.

## Addressed

**HIGH**
- **H1** — `Sdk.targets` regex parity. Both PropertyGroups now mirror all 5 patterns from `RepoUrlBuilder.cs`: GitHub HTTPS (creds-tolerant), GitHub SSH (covers `git@github.com:` + `ssh://git@github.com/` via `[:/]`), ADO modern HTTPS, ADO modern SSH (`git@ssh.dev.azure.com:v3/...`), and ADO legacy HTTPS (`{org}.visualstudio.com`). No new task assembly — still inline `Regex.Match` MSBuild property functions.
- **H2** — `SecurityException` parity at the four sites that were missing it (`SourceMapFileHandler.HandleAsync` `Path.GetFullPath` + `FileInfo`, `SourceMapSources.TryParseAsync` `FileInfo`). `TryGetFileSize` rounded out with `ArgumentException` / `PathTooLongException` / `NotSupportedException` so the 4-exception set is consistent across the codebase.
- **H3** — direct unit tests for `TryBuildRepoRedirect`. Covers hostile schemes (`javascript:`, `data:`, `file://`, `vbscript:`, `ftp://`), null/empty inputs, the `sources[]` membership gate, http-vs-https split, and the new path encoding.

**MEDIUM**
- **M1** — gate map parsing on `ServeFromSourcesLong || RepoUrlRedirectOnMiss`. Default-config requests skip file I/O + JSON parse entirely.
- **M2** — wrap legacy `Path.GetFullPath` fallback in `SourceMap.ToString` (extracted as `AbsolutizeForSources`) with the 4-exception set, degrading to the unprocessed forward-slashed form on failure.
- **M3** — `TryRebaseToRepoRoot` now uses `OperatingSystem.IsWindows()` to pick the comparer, matching `IsContained` and avoiding case-collision false positives on Linux/macOS.
- **M4** — direct unit tests for `NormalizeRepoRoot`, `TryRebaseToRepoRoot`, and `SourceMapSources.SourceRoot`. Internals exposed via `[assembly: InternalsVisibleTo(\"SourceMap.Test\")]` declared in source (the csproj has `GenerateAssemblyInfo=false`, so the MSBuild item form does not flow through).

**LOW**
- **L1** — `TryBuildRepoRedirect` now refuses `http://` in addition to non-http(s); both GitHub raw and the ADO Items API are HTTPS-only.
- **L2** — redirect URL builder URI-encodes path segments via `Uri.EscapeDataString`, preserving `/` as a separator. A `#` in a source name no longer truncates the ADO `path=/` query at the fragment boundary.

## Deferred (nits, not addressed)

- **L3** and **L4** — noted as nits in the iter-1 review, deferred for a future pass.

gautam-achieveai (commented):
## [Gautam's review bot] Iteration-2 Re-review — ✅ APPROVE

### Context
**User Story #19**: Support linking source maps to ADO/GitHub repository files.
Incremental review of `19961501` vs previously reviewed `63073ab`. 7 files changed, +541/−18.

### Verification
- **SourceMap.Test:** 85/85 pass ✅
- **SourceMap.Server.Test:** 60/60 pass ✅
- **Agents used:** Security-focused code review, manual deep review

---

### Iteration-1 Findings — All Addressed

| # | Sev | Finding | Status |
|---|-----|---------|--------|
| H1 | 🔴 HIGH | Sdk.targets regex misses 3/5 URL variants | ✅ Fixed — 5 named patterns now match `RepoUrlBuilder.cs` |
| H2 | 🔴 HIGH | Missing `SecurityException` at 4 sites | ✅ Fixed — all 4 sites + `TryGetFileSize` rounded out |
| H3 | 🔴 HIGH | `TryBuildRepoRedirect` untested for hostile schemes | ✅ Fixed — 14 new direct test methods |
| M1 | 🟡 MED | Unnecessary map parsing when both features off | ✅ Fixed — `needsParse` gate |
| M2 | 🟡 MED | Unprotected `Path.GetFullPath` fallback | ✅ Fixed — `AbsolutizeForSources` with 4-exception wrapping |
| M3 | 🟡 MED | `OrdinalIgnoreCase` on case-sensitive FS | ✅ Fixed — platform-aware comparer |
| M4 | 🟡 MED | Missing direct tests for internal helpers | ✅ Fixed — `InternalsVisibleTo` + 11 new tests |
| L1 | 🔵 LOW | `http://` allowed in redirect | ✅ Fixed — `https://` only |
| L2 | 🔵 LOW | Missing URL encoding on redirect path | ✅ Fixed — bonus `EncodeSourcePath` |

### New Observations (Non-blocking, Informational)

**1. `EncodeSourcePath` — well-done bonus improvement**
Segment-level URI encoding correctly prevents `#`, `?`, `&`, space from corrupting ADO query strings. Good proactive hardening not requested in iter-1.

**2. Defense-in-depth consideration (future work)**
`TryBuildRepoRedirect` validates `https://` scheme + `sources[]` membership — both solid. For additional hardening consider:
- Rejecting source names containing `..` path segments (prevents cross-repo traversal via malicious `.map`)
- Optional domain allowlist for `sourceRoot` (`raw.githubusercontent.com`, `dev.azure.com`)

Low-probability scenarios (require attacker to control `.map` file), but would tighten the security boundary further.

### Overall
All 8 actionable findings properly addressed with clean implementations and test coverage. Exception handling convention now consistent. No new blocking issues. Ship it. 🚢

Sources/Compiler/NScript.Sdk/Sdk/Sdk.targets
gautam-achieveai: [Gautam's review bot] ✅ **Verified in iter-2.** All 5 patterns now present with cascading Condition guards. GitHub HTTPS/SSH and ADO modern HTTPS/SSH/legacy HTTPS all covered. Comments reference RepoUrlBuilder.cs for sync. No remaining parity gaps.

Sources/Compiler/SourceMap.Server/SourceMapFileHandler.cs
gautam-achieveai: [Gautam's review bot] ✅ **Verified.** Line 364 now rejects anything other than https://. Test TryBuildRepoRedirect_HttpSourceRoot_ReturnsFalse explicitly covers plain http://.

Sources/Compiler/SourceMap/SourceMap.cs
gautam-achieveai: [Gautam's review bot] ✅ **Verified.** OperatingSystem.IsWindows() selects between OrdinalIgnoreCase (Windows) and Ordinal (Linux/macOS), matching the IsContained pattern in SourceMapFileHandler.

Sources/Compiler/SourceMap.Server/SourceMapFileHandler.cs
gautam-achieveai: [Gautam's review bot] ✅ **Verified.** SecurityException catches added at all 4 sites: HandleAsync Path.GetFullPath (L184), HandleAsync new FileInfo (L220), SourceMapSources.TryParseAsync (L114), TryGetFileSize (L434). Convention now consistent across all file-system access points.

Sources/Compiler/SourceMap.Server/SourceMapFileHandler.cs
gautam-achieveai: [Gautam's review bot] ✅ **Verified.** needsParse gate at L137 skips TryParseAsync when both ServeFromSourcesLong and RepoUrlRedirectOnMiss are false. Default-config requests skip all disk I/O.

Sources/Compiler/SourceMap/SourceMap.cs
gautam-achieveai: [Gautam's review bot] ✅ **Verified.** AbsolutizeForSources cleanly extracts the fallback with 4-exception wrapping. On failure, degrades to rawFile.Replace("\\\\", "/") — correct for the escaped form from AddMapping. Minor note: fallback omits ":" → "$" replacement, acceptable for malformed-path degradation.

gautam-achieveai:
[Gautam's Dev bot]

Thanks for the iter-2 ✅. Acknowledging the two non-blocking observations as future-work follow-ups (will not address in this PR per the "Ship it" disposition):

1. **Reject `..` path segments in `TryBuildRepoRedirect`** — additional defense-in-depth alongside the existing `sources[]` membership gate. Cheap to add; tracking for a follow-up.
2. **Optional `sourceRoot` domain allowlist** (`raw.githubusercontent.com`, `dev.azure.com`) — fits as an opt-in option on the handler config alongside `RepoUrlRedirectOnMiss`. Tracking for a follow-up.

Disposition: deferred to follow-up issue.

PR #46, opened 2026-04-26
[Gautam's Dev bot] WI-38: WebStoreTable resumable pagination cursor tokens

### LINKED ISSUE #38: WebStoreTable: add resumable pagination cursor tokens
[Gautam's Dev bot]

## Context

Follow-up tracking issue from #34.

Issue #34 addresses the immediate problem of the silent `1 << 20` default row cap by:

1. Removing the implicit cap.
2. Adding a streaming `ForEach` overload for \"visit every record without materialising\" use cases.
3. Rejecting accidental unbounded `Query(Query.All)` materialisation at the read path.

Resumable pagination cursor tokens were explicitly marked out of scope for #34 because they add significant API surface area (serializable cursor state, resume semantics, ordering guarantees across transactions) that none of the current callers require.

This issue tracks that future work.

## Proposed scope (sketch — to be refined when prioritised)

- Add an opaque `Cursor` / `PaginationToken` type returned from `WebStoreTable.QueryPage(Query, int pageSize, Cursor resumeFrom = null)`.
- Cursor encodes last-seen primary key + direction so subsequent `QueryPage` calls resume deterministically.
- Token must be serializable (string or byte[]) so callers can persist across sessions.
- Define ordering/consistency guarantees when records are mutated between page fetches.
- Add tests for: first page, subsequent pages, mutation between pages, exhaustion.

## Why not now

- No current caller in the repo needs resumable pagination — streaming `ForEach` (delivered in #34) covers all in-tree use cases.
- API shape deserves its own design pass rather than being bundled with the cap-removal refactor.

## Dependencies

- Blocked on #34 landing (establishes the post-cap baseline and introduces the `ForEach` streaming primitive that cursor pagination will sit alongside).

gautam-achieveai:
[Gautam's Dev bot] Closes #38.

## Summary

Adds resumable, bounded-size pagination to `WebStoreTable` via a new `QueryPage` / `QueryKeysPage` API. Callers get a `Page<TValue>` (items capped at pageSize) with an opaque `Cursor` continuation token they pass back to fetch the next page.

## What changed

**New types** (`Sources/Framework/Sunlight.Framework.Data/WebStore/`):
- `Page<TValue>` — `Items` (post-filter, capped at pageSize) + `NextCursor` (null when the scan drained before the page filled).
- `Cursor` — opaque continuation token carrying `(IndexKey, PrimaryKey, Direction, TableSig)`. Round-trips via `ToToken` / `FromToken` using a literal-keyed JSON envelope (`v`/`d`/`t`/`p`/`i`) so NScript minification cannot break the wire format across builds.

**WebStoreTable additions**:
- `QueryPage(query, pageSize, resumeFrom?, inclFilter?, transaction?)` and `QueryKeysPage(query, pageSize, resumeFrom?, transaction?)`.
- Private `QueryPageInternal<U>` validates `pageSize >= 1`, rejects `Skip`/`Limit` when resuming, enforces table-sig and direction match, then drives the existing `CursorIterator` with skip-until-past-token logic. Index ties tiebreak on primary key per IDB-spec ordering.

## Semantics

- **pageSize** counts records that pass the filter — "I want N rows", not "N reads".
- **NextCursor** is `null` only when the iterator drained the scan range before the page filled; non-null signals at least one more record may exist.
- **See-current-state** mutation contract: rows inserted strictly past the cursor become visible on the next page; rows inserted before are silently missed; deletes are skipped naturally.

## Validation surface (all reject through the Promise — symmetric with rest of the API)

- `pageSize <= 0`
- `resumeFrom` combined with `Query.Skip` or `Query.Limit`
- Cursor issued for a different table (`TableSig` mismatch)
- Cursor direction doesn't match the query's direction

## Tests

14 QUnit tests in `WebStoreCrudTests.cs` cover the full API surface:
- First page returns `pageSize` rows + non-null cursor
- Resume produces non-overlapping continuation
- End-of-stream returns null cursor
- Empty table resolves with empty page
- Token round-trip preserves continuation
- Descending direction
- Non-unique index with ties
- Mutation past cursor visible / before cursor missed (see-current-state)
- Filter counts accepted rows, not reads
- Rejection paths: `pageSize <= 0`, cross-table token, direction mismatch
- Keys-only variant

All 230 QUnit tests pass; both Debug and Release builds succeed (0 errors).

## Test plan

- [x] `dotnet build NScript_Full.sln -c Debug` — 0 errors
- [x] `dotnet build NScript_Full.sln -c Release` — 0 errors
- [x] QUnit suite via `Test/Framework/TestWebApplication/run-qunit.mjs` — 230/230 pass

### DIFF
--- a/Sources/Framework/Sunlight.Framework.Data/WebStore/Cursor.cs
+++ b/Sources/Framework/Sunlight.Framework.Data/WebStore/Cursor.cs
@@ -78,8 +78,8 @@ namespace Sunlight.Framework.Data.WebStore
 
             Dictionary env;
             try { env = JSON.Parse(token); }
-            catch (Exception)
-            { throw new Exception("Cursor token is not valid JSON"); }
+            catch (Exception ex)
+            { throw new Exception("Cursor token is not valid JSON: " + ex.Message); }
 
             if (env == null)
             { throw new Exception("Cursor token decoded to null"); }
--- a/Sources/Framework/Sunlight.Framework.Data/WebStore/Cursor.cs
+++ b/Sources/Framework/Sunlight.Framework.Data/WebStore/Cursor.cs
@@ -1,0 +1,112 @@
+namespace Sunlight.Framework.Data.WebStore
+{
+    using System;
+    using System.Collections;
+    using System.Web;
+
+    /// <summary>
+    /// Opaque continuation token issued by
+    /// <see cref="WebStoreTable{TKey,TValue}.QueryPage"/> /
+    /// <see cref="WebStoreTable{TKey,TValue}.QueryKeysPage"/>. Round-trips to a
+    /// string via <see cref="ToToken"/> / <see cref="FromToken(string)"/> so
+    /// callers can persist it across sessions, network hops, or storage tiers
+    /// without reaching into the cursor's internal state.
+    /// <para>
+    /// The wire format is a JSON envelope with literal short keys (<c>v</c>,
+    /// <c>d</c>, <c>p</c>, <c>i</c>, <c>t</c>) — NScript minification renames
+    /// C# field names, so reflecting on field names would break the round-trip
+    /// across builds.
+    /// </para>
+    /// </summary>
+    public sealed class Cursor
+    {
+        private const string TokenVersion = "1";
+
+        internal Cursor(
+            object indexKey,
+            object primaryKey,
+            string direction,
+            string tableSig)
+        {
+            IndexKey = indexKey;
+            PrimaryKey = primaryKey;
+            Direction = direction;
+            TableSig = tableSig;
+        }
+
+        /// <summary>
+        /// Index key of the last accepted record, or <c>null</c> when the cursor
+        /// was issued by a primary-key scan (no secondary index involved).
+        /// </summary>
+        internal object IndexKey { get; }
+
+        /// <summary>Primary key of the last accepted record. Always populated.</summary>
+        internal object PrimaryKey { get; }
+
+        /// <summary>IDB cursor direction string ("next" or "prev") at issue time.</summary>
+        internal string Direction { get; }
+
+        /// <summary>Table-signature (object-store name) at issue time.</summary>
+        internal string TableSig { get; }
+
+        /// <summary>
+        /// Encode the cursor into a stable, persistable string. The encoding is
+        /// stable across NScript minified builds because the JSON keys are
+        /// literal — never C# field names.
+        /// </summary>
+        public string ToToken()
+        {
+            var env = new Dictionary();
+            env.Set("v", TokenVersion);
+            env.Set("d", Direction);
+            env.Set("t", TableSig);
+            env.Set("p", PrimaryKey);
+            if (IndexKey != null)
+            { env.Set("i", IndexKey); }
+            return JSON.Stringify(env);
+        }
+
+        /// <summary>
+        /// Decode a cursor from a token previously produced by
+        /// <see cref="ToToken"/>. Throws a descriptive <see cref="Exception"/>
+        /// when the token is malformed or carries an unsupported version.
+        /// </summary>
+        public static Cursor FromToken(string token)
+        {
+            if (token == null || token.Length == 0)
+            { throw new Exception("Cursor token is null or empty"); }
+
+            Dictionary env;
+            try { env = JSON.Parse(token); }
+            catch (Exception)
+            { throw new Exception("Cursor token is not valid JSON"); }
+
+            if (env == null)
+            { throw new Exception("Cursor token decoded to null"); }
+
+            if (!env.ContainsKey("v") || env.Get<string>("v") != TokenVersion)
+            { throw new Exception("Cursor token version mismatch (expected v=1)"); }
+
+            if (!env.ContainsKey("d"))
+            { throw new Exception("Cursor token missing direction"); }
+
+            var direction = env.Get<string>("d");
+            if (direction != "next" && direction != "prev")
+            { throw new Exception("Cursor token has invalid direction"); }
+
+            if (!env.ContainsKey("t"))
+            { throw new Exception("Cursor token missing table signature"); }
+
+            var tableSig = env.Get<string>("t");
+            if (tableSig == null || tableSig.Length == 0)
+            { throw new Exception("Cursor token missing table signature"); }
+
+            if (!env.ContainsKey("p"))
+            { throw new Exception("Cursor token missing primary key"); }
+
+            var primaryKey = env.Get<object>("p");
+            var indexKey = env.ContainsKey("i") ? env.Get<object>("i") : null;
+            return new Cursor(indexKey, primaryKey, direction, tableSig);
+        }
+    }
+}
--- a/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
+++ b/Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
@@ -241,6 +241,74 @@ namespace Sunlight.Framework.Data.WebStore
                     true));
         }
 
+        /// <summary>
+        /// Materialise a single bounded page of records matching
+        /// <paramref name="query"/>. Pagination semantics:
+        /// <list type="bullet">
+        /// <item><description><paramref name="pageSize"/> must be &ge; 1 and counts
+        /// records that pass <paramref name="inclFilter"/> — "I want N rows", not
+        /// "N reads". The returned <see cref="Page{TValue}.Items"/> is capped at
+        /// <paramref name="pageSize"/>.</description></item>
+        /// <item><description><see cref="Page{TValue}.NextCursor"/> is <c>null</c>
+        /// when the iterator drained the scan range before the page filled — that
+        /// page is the final page. A non-<c>null</c> cursor signals at least one
+        /// additional record may exist.</description></item>
+        /// <item><description>Pass <paramref name="resumeFrom"/> as the previous
+        /// page's <see cref="Page{TValue}.NextCursor"/> to fetch the next page.
+        /// Cross-transaction <i>see-current-state</i> semantics: records inserted
+        /// strictly past the cursor become visible on the next page; records
+        /// inserted strictly before are silently missed (cursor moved past);
+        /// deleted records are skipped naturally as the cursor walks current
+        /// state.</description></item>
+        /// </list>
+        /// Rejects when <paramref name="pageSize"/> &le; 0, when
+        /// <paramref name="resumeFrom"/> is combined with <c>Query.Skip</c> or
+        /// <c>Query.Limit</c>, when the cursor was issued for a different table,
+        /// or when its direction doesn't match <paramref name="query"/>'s.
+        /// </summary>
+        public Promise<Page<TValue>> QueryPage(
+            Query query,
+            int pageSize,
+            Cursor resumeFrom = null,
+            Func<TValue, bool> inclFilter = null,
+            WebStoreTransaction transaction = null)
+        {
+            return new Promise<Page<TValue>>((resolve, reject) =>
+                this.QueryPageInternal<TValue>(
+                    transaction,
+                    query,
+                    pageSize,
+                    resumeFrom,
+                    inclFilter,
+                    resolve,
+                    reject,
+                    false));
+        }
+
+        /// <summary>
+        /// Key-only paged variant of <see cref="QueryPage"/> — returns primary
+        /// keys instead of values. Same pagination semantics; same validation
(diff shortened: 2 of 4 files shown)


gautam-achieveai (commented):
## [Gautam's review bot] PR !46 — Deep Review

**Linked:** #38 (WebStoreTable: add resumable pagination cursor tokens)

---

### 🟢 Verdict: Approve with Comments

Well-implemented feature with clean architecture, solid test coverage (14 tests), and thorough documentation. The `Cursor`/`Page<T>`/`QueryPage` separation is well-structured with appropriate visibility boundaries. Literal JSON keys correctly survive NScript minification. A few items warrant attention — one performance concern for follow-up and a couple of test gaps.

---

### Architecture (✅ Clean)
- `Cursor` → owns serialization format with internal-only construction. `Page<T>` → minimal container. `QueryPageInternal` → all iteration logic. Good separation of concerns.
- Follows established `public wrapper → private Internal<U>` pattern used by `Query`/`QueryKeys`/`ForEach`.
- WebStoreTable at 1242 lines is growing but this PR is a natural, cohesive extension of the existing query axis. Not blocking; consider partial-class splitting if more query variants follow.

### Review Agents Used
`architecture-review`, `performance-review`, `exception-handling-review`, `test-coverage-review`

---

### Findings

#### 1️⃣ [HIGH] O(N) resume skip per page — O(N²) total pagination cost
**`WebStoreTable.cs:979-1008`**

When resuming, `CursorIterator` opens from the **start** of the key range and walks record-by-record (`cursor.Continue()`) until past the cursor position. For page M with pageSize P, this walks ~(M-1)×P records before collecting anything.

**Concrete impact:** 10,000 records / pageSize=50 → page 200 walks ~10,000 cursor steps (~100-500ms of main-thread IDB work). Total across all 200 pages: ~1M cursor steps vs. optimal 10K.

**The fix is already available:** `IDBCursor.Continue(IDBKeyRange key)` (bound at `IDBCursor.cs:37`) does an O(log N) B-tree seek. It parallels how the existing `CursorIterator` already handles `Skip` via `cursor.Advance(skip)` at line 600-606. Adding an optional initial-seek key to `CursorIterator` would reduce per-page resume from O(N) to O(log N).

**Not blocking this PR** — the API is functionally correct, and the optimization requires modifying the shared `CursorIterator`. But this should be tracked as a follow-up work item, especially if tables with >1K records are expected.

#### 2️⃣ [HIGH] Filter + resume interaction untested
**`WebStoreCrudTests.cs`**

Test 10 (`TestQueryPageWithFilterSkipsRejectedRows`) exercises filter on first page only — never resumes with the returned cursor. The `CursorIterator` applies `inclFilter` *before* the resume-skip callback sees the row (line 609). If cursor position interacted incorrectly with pre-filtering, it would be invisible.

**Suggested test:** Seed 10 alternating-category rows, fetch page 1 with `inclFilter` accepting `"todo"` + `pageSize=2`, resume with cursor + same filter, assert page 2 contains 2 `"todo"` rows with no overlap.

#### 3️⃣ [MEDIUM] Skip + cursor guard untested
**`WebStoreTable.cs:937-942`**

This rejection guard exists but no test exercises it. Without it, `Skip` would silently create gaps in paginated results — nearly impossible to diagnose from the caller's side.

**Suggested test:** Build query with `Skip(2)`, get cursor from page 1, call `QueryPage` with both skip-query and cursor, assert rejection with message containing `"Skip"`.

#### 4️⃣ [MEDIUM] Limit + cursor guard untested
**`WebStoreTable.cs:944-949`**

Same gap. Without this guard, `Limit` would cap total reads below `pageSize`, causing premature short pages that callers interpret as end-of-stream.

#### 5️⃣ [MEDIUM] Inner exception swallowed in `FromToken`
**`Cursor.cs:81-82`**

```csharp
try { env = JSON.Parse(token); }
catch (Exception)
{ throw new Exception("Cursor token is not valid JSON"); }
```

The original parse error (offset, unexpected character) is discarded. Easy fix:
```csharp
catch (Exception ex)
{ throw new Exception("Cursor token is not valid JSON: " + ex.Message); }
```

---

### Positive Observations
- ✅ Literal JSON keys (`"v"`, `"d"`, `"p"`, `"i"`, `"t"`) correctly survive NScript minification — well-reasoned design for a cross-build wire format
- ✅ `internal` constructor + properties on `Cursor` form clean trust boundary — only `WebStoreTable` can create/inspect cursors
- ✅ All 5 validation guards in `QueryPageInternal` properly use `reject()` + `return` (no fall-through)
- ✅ `CursorIterator`'s try-catch blanket properly catches `ComparePagePair` exceptions during iteration
- ✅ XML documentation is comprehensive — especially the see-current-state semantics on `QueryPage`
- ✅ Test quality is high with behavioral assertions (item content, ordering, overlap), not mock-based

Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
```diff
@@ -850,6 +918,139 @@ private void QueryInternal<U>(
...
+                        int cmp = ComparePagePair(
+                            isIndex,
+                            curIndexKey,
+                            curPrimaryKey,
+                            resumeFrom.IndexKey,
+                            resumeFrom.PrimaryKey);
+                        bool past = descending ? cmp < 0 : cmp > 0;
+                        if (!past)
+                        { return true; }
+                        resumed = true;
```
gautam-achieveai, line 1008: **[Gautam's review bot] [HIGH] O(N) resume skip — performance concern**

`CursorIterator` opens from the start of the key range every time. During resume, this walks (M-1)×P records before collecting anything for page M. Total pagination cost is O(N²×P).

`IDBCursor.Continue(IDBKeyRange key)` is already bound in the codebase (`IDBCursor.cs:37`) and does O(log N) B-tree seek — the same pattern as `cursor.Advance(skip)` at line 600. Adding an optional initial-seek key to `CursorIterator` would eliminate the linear skip.

Not blocking, but recommend tracking as a follow-up work item — later pages on tables with >1K records will be noticeably slow.

Sources/Framework/Sunlight.Framework.Data/WebStore/Cursor.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// </summary>
+        public static Cursor FromToken(string token)
+        {
+            if (token == null || token.Length == 0)
+            { throw new Exception("Cursor token is null or empty"); }
+
+            Dictionary env;
+            try { env = JSON.Parse(token); }
+            catch (Exception)
+            { throw new Exception("Cursor token is not valid JSON"); }
```
gautam-achieveai: **[Gautam's review bot] [MEDIUM] Inner exception swallowed**

The original `JSON.Parse` exception carries diagnostic details (parse offset, unexpected character) that are discarded here.

Suggested fix:
```csharp
catch (Exception ex)
{ throw new Exception("Cursor token is not valid JSON: " + ex.Message); }
```

Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
```diff
@@ -850,6 +918,139 @@ private void QueryInternal<U>(
...
+                reject(new Exception("pageSize must be >= 1"));
+                return;
+            }
+
+            if (resumeFrom != null && query.Skip != null)
+            {
+                reject(new Exception(
+                    "Skip and resume cursor are mutually exclusive"));
+                return;
+            }
```
gautam-achieveai, line 942: **[Gautam's review bot] [MEDIUM] Skip + cursor and Limit + cursor guards untested**

These two rejection guards (lines 937-949) are the only validation branches in `QueryPageInternal` without test coverage. The other three (pageSize, table sig, direction) are pinned by tests 11–13.

Without tests, a refactor could remove these silently:
- **Skip + cursor:** Would create gaps in paginated results (Skip applied on top of resume)
- **Limit + cursor:** Would cause premature short pages interpreted as end-of-stream, silently truncating results

Suggested: add quick rejection tests for both paths.

Sources/Framework/Sunlight.Framework.Data/WebStore/Cursor.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 44f724b1 — `FromToken` now preserves the inner exception message:

```csharp
catch (Exception ex)
{ throw new Exception("Cursor token is not valid JSON: " + ex.Message); }
```

Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
gautam-achieveai: [Gautam's Dev bot]

Added in 44f724b1:

- `TestQueryPageRejectsSkipWithCursor` — asserts rejection when `QueryBuilder.Skip(1)` is paired with a resume cursor; rejection message contains "Skip".
- `TestQueryPageRejectsLimitWithCursor` — asserts rejection when `QueryBuilder.Limit(3)` is paired with a resume cursor; rejection message contains "Limit".

Sources/Framework/Sunlight.Framework.Data/WebStore/WebStoreTable.cs
gautam-achieveai: [Gautam's Dev bot]

Acknowledged — agreed this is a valid follow-up. The suggested fix (threading an optional initial-seek `IDBKeyRange` into `CursorIterator` and using `cursor.Continue(keyRange)` for the first hop) is correct and would convert resume from O(N) to O(log N).

Tracking as a separate work item rather than expanding this PR — the change touches the shared `CursorIterator` (used by `Query`, `QueryKeys`, `ForEach`, plus the new paging path) and warrants its own performance-focused review and benchmark coverage.

gautam-achieveai:
[Gautam's Dev bot]

Addressing the review feedback in 44f724b1:

| Finding | Severity | Resolution |
|---|---|---|
| #1 — O(N) resume skip | HIGH | Acknowledged as follow-up (touches shared `CursorIterator`); replied inline. |
| #2 — Filter + resume untested | HIGH | Added `TestQueryPageFilterResumeIsContiguous` — fetches page 1 with `inclFilter=todo`, resumes with same filter, asserts contiguous non-overlapping `r0,r2,r4,r6` sequence. |
| #3 — Skip + cursor guard untested | MEDIUM | Added `TestQueryPageRejectsSkipWithCursor`. |
| #4 — Limit + cursor guard untested | MEDIUM | Added `TestQueryPageRejectsLimitWithCursor`. |
| #5 — Inner exception swallowed | MEDIUM | Fixed in `Cursor.FromToken` — now appends inner `JSON.Parse` `ex.Message`. |

Debug build: clean (0 errors). Three inline comments replied to individually.

gautam-achieveai (commented):
## [Gautam's review bot] PR !46 — Incremental Review (Iteration 2)

**Linked:** #38 (WebStoreTable: add resumable pagination cursor tokens)

---

### ✅ Verdict: Approve

All 5 findings from the first review have been addressed cleanly in `44f724b`. No new issues introduced.

---

### First-Review Findings — Resolution Status

| # | Finding | Severity | Resolution | Verified |
|---|---------|----------|------------|----------|
| 1 | O(N) resume skip — O(N²) total pagination cost | HIGH | Developer acknowledged; tracking as follow-up work item (correct call — touches shared `CursorIterator`) | ✅ |
| 2 | Filter + resume interaction untested | HIGH | Added `TestQueryPageFilterResumeIsContiguous` — alternating "todo"/"done" rows, verifies page 2 resumes at r4 with no overlap or gap | ✅ |
| 3 | Skip + cursor guard untested | MEDIUM | Added `TestQueryPageRejectsSkipWithCursor` — asserts rejection with "Skip" in message | ✅ |
| 4 | Limit + cursor guard untested | MEDIUM | Added `TestQueryPageRejectsLimitWithCursor` — asserts rejection with "Limit" in message | ✅ |
| 5 | Inner exception swallowed in `FromToken` | MEDIUM | Fixed: `catch (Exception ex) { throw new Exception("... " + ex.Message); }` | ✅ |

### Quality of New Tests

All 3 new tests are well-crafted:

- **Rejection tests** (Skip/Limit) follow the established promise-rejection pattern (`Then<bool, object>` with resolve-fail + reject-assert). Consistent with existing rejection tests for pageSize, cross-table, and direction mismatch.
- **Filter+resume contiguity test** is the standout — the alternating-category seed data forces the exact interaction between `inclFilter` and the resume-skip predicate. Assertions verify observable behavior (item IDs in order), not implementation details.

### Review Agents Used
`test-coverage-review`, `exception-handling-review`

### Open Follow-Up
- Performance optimization for cursor resume (O(N) → O(log N) via `IDBCursor.Continue(key)`) — acknowledged by developer, to be tracked as separate work item.

PR #52, opened 2026-05-04
WI-51: split QUnit pages so each loads only one *.Test.js bundle

### LINKED ISSUE #51: Multi-module load corrupts Function.prototype runtime metadata (per-IIFE letter collision)
## Summary

When two or more NScript-generated JS files are loaded together in one HTML page (e.g., `Sunlight.Framework.Test.js` + `Sunlight.Framework.UI.Test.js` + `TodoApp.Test.js` in `Test/Framework/TestWebApplication/TestPage.htm`), every module after the first has its runtime type-system slots on `Function.prototype` silently clobbered. This produces wrong-looking `TypeError`s such as `d.b_h is not a function` or `InvalidCast to ...` in tests that pass when run alone.

## Reproduction (current master)

```bash
cd Test/Framework/TestWebApplication
npx serve .
# open http://localhost:3000/TestPage.htm in a browser, observe 8 failures including:
#   DataBinderTests.BasicBinderSimpleObjectTest, BasicBinderSimpleObjectTestReverseOrder
#   DataBinderTests.BasicBinderOneWayNotifiableObjectTest, BasicBinderTwoWayNotifiableObjectTest
#   SourcePropertyBinderTests.BasicValueTest, BasicValueTestWithNotification, PropertyPathValueTest, PropertyPathValueNotifiableTest
```

If only `Sunlight.Framework.Test.js` is loaded (without UI.Test or TodoApp.Test), all those tests pass.

If you reverse the script-tag order (TodoApp.Test.js first, Sunlight.Framework.Test.js last), the same kind of bug shifts: now Sunlight.Framework.Test passes, but **87** tests in `Sunlight.Framework.UI.Test.LiveBinderTests` etc. fail. Whichever module is NOT last gets clobbered.

## Root cause

Every NScript-compiled JS file is wrapped in its own IIFE and minified independently. The runtime metadata sentinels written onto `Function.prototype` (is-class flag, is-interface flag, type-check method, FQN field, interface map field, default-ctor stub, etc.) are renamed to short letters by the per-module minifier — and the minifier picks DIFFERENT letters per module.

Letter mappings observed in the three current generated files:

| Concept | `Sunlight.Framework.Test.js` | `Sunlight.Framework.UI.Test.js` | `TodoApp.Test.js` |
|---|---|---|---|
| is-class flag | `r=true` | `H=true` | `s=true` |
| is-interface flag | `p=true` | `z=true` | `r=true` |
| type-check method | `o = function(a){...}` | `m = function(a){...}` | `p = function(a){...}` |
| no-op stub assigned to letter `o` | (real type-check) | (no `o` assigned) | `o = function(a){return null;}` |
| FQN string field | `e` | `c` | `b` |
| interface map field | `h` | `l` | `h` |

Because each IIFE writes `Function.prototype.<letter> = ...` for its own letter mapping, but they all share the same global `Function.prototype`, the LAST module's mapping wins. Examples of the resulting collisions:

- After `TodoApp.Test.js` loads, `Function.prototype.o` = `function(a){return null;}` (a no-op stub in TodoApp's letter mapping). But `Sunlight.Framework.Test.js`'s cast helper does `b.o(a)` expecting the real type-check — so every cast through that path gets the wrong answer.
- After `Sunlight.Framework.Test.js` loads (when reordered last), `Function.prototype.m = false` (from its defaults-init block). But `Sunlight.Framework.UI.Test.js`'s cast helper does `b.m(a)` because `m` was UI.Test's type-check letter — so `b.m is not a function` follows.
- The is-interface boolean flag uses different letters per module, so a class registered in module A may end up with a function value (rather than a boolean) in the slot that module B's type-check reads, flipping the interface-vs-class branch and producing `d.b_h is not a function` (the interface-method dispatch goes down the wrong path on a value that doesn't have that method).

## Why this is invisible to the pre-commit hook

The pre-commit hook runs only the TodoApp e2e flow via `e2e-todo-tests.js`, which loads only TodoApp scripts on its page. The QUnit Binders suite is never exercised by the hook, so the hook reports green even with the failures present.

## Where the fix likely lives

This is a runtime-metadata-naming bug in NScript codegen. Options:

1. **Reserve the metadata letters** so the minifier picks the same names across all modules. (Hard if minification is independent; would require coordinating a shared symbol table.)
2. **Use stable, non-minified names** for the runtime metadata slots — e.g., `Function.prototype.__nscript_typeCheck__`, `__nscript_isInterface__`, `__nscript_fqn__`, etc. These are written once per module but always to the same key. (Simplest fix; small constant size cost.)
3. **Use `Symbol`-based property keys** for runtime metadata, generated/exported by a single shared bootstrap module that every generated module imports. Avoids string collisions entirely.
4. **Single shared bootstrap script**: extract the `Function.prototype` runtime contract into a separate JS file loaded ONCE per page, and have generated modules consume those slots rather than installing their own.

Option (2) is probably the smallest, most local change.

## Codebase pointers (suggested starting points)

- `Sources/Compiler/NScript.JST` — JavaScript AST + minification.
- The runtime emitter that writes the `Function.prototype` defaults block (search generated output for `a=Function.prototype;a.m=false;` to find the source emitter).
- `feedback_nscript_identifier_resolution` memory and the existing CLAUDE.md note: "ALL identifiers must be resolved via scope system, never raw strings." The runtime contract on `Function.prototype` is exactly the kind of stable identifier set that must not be subject to per-module renaming.

## Relationship to PR #50

PR #50 only bumps NuGet package versions; it does not touch codegen or runtime. This bug exists on master before #50 and is independent of the package bumps.

gautam-achieveai:
[Gautam's Dev bot]

## Summary

Closes #51

Each NScript-compiled bundle is its own IIFE minified independently and writes runtime metadata (type-check, is-class flag, FQN, interface map, …) to short letters on `Function.prototype`. Co-loading two bundles on one page silently clobbers earlier metadata with the next bundle's letter mapping, producing `b.m is not a function` / `InvalidCast` errors in random downstream tests (issue #51).

`DataTestPage.htm` already isolates one bundle per page for exactly this reason. This PR extends the same pattern to the remaining three suites:

- Add `Sunlight.Framework.TestPage.htm` — loads only `Sunlight.Framework.Test.js`
- Add `Sunlight.Framework.UI.TestPage.htm` — loads only `Sunlight.Framework.UI.Test.js`
- Add `TodoApp.TestPage.htm` — loads only `TodoApp.Test.js`
- Convert `TestPage.htm` into a no-bundle **index page** linking to the four suites
- Update `run-tests.js` to iterate the four per-suite pages and aggregate pass/fail counts
- Register the new pages as `<Content Include>` in `TestWebApplication.csproj`
- Reword `Test/CLAUDE.md`, `Test/docs/browser-testing.md`, `docs/testing/README.md` to point at the per-suite pages and explain why bundles are not co-loaded

`run-qunit.mjs` already isolates one bundle per page (it synthesises a one-bundle page per suite via `/__qunit?name=`) and is unchanged. The pre-commit e2e flow (`TodoApp.htm` + `e2e-todo-tests.js`) is unaffected.

## Scope (and what this is NOT)

This is the **avoidance** fix from the plan approved on issue #51. It does **not** fix the root cause (per-IIFE letter collision on `Function.prototype`). The four candidate root-cause fixes listed in the issue (stable non-minified property names, `Symbol`-based keys, single shared bootstrap, or coordinating the minifier across bundles) all change the runtime ABI for `Function.prototype` slot access and are materially larger. **Recommendation:** open a follow-up issue tracking the root-cause fix (Option 2 from the bug report — stable `__nscript_*__` property names — is still the smallest viable patch). The avoidance fix here doesn't preclude the root-cause fix later.

## Test plan

- [ ] CI Phase 3 (`run-qunit.mjs`) passes on all four suites — this is the primary regression guard.
- [ ] CI Phase 3 (`e2e-todo-tests.js`) still passes — `TodoApp.htm` and the e2e bundle layout are untouched.
- [ ] `node run-tests.js` (after change) iterates the four per-suite pages and reports aggregate counts.
- [ ] Manual smoke (local): open each new `*.TestPage.htm`; QUnit banner reaches `qunit-pass`. Open the new index `TestPage.htm`; verify it loads no bundles and the four links resolve.
- [ ] Negative regression: confirm that on master before this change, opening the OLD multi-bundle `TestPage.htm` reproduces the 8 failures cited in the issue (`DataBinderTests.BasicBinderSimpleObjectTest`, etc.); then confirm those failures disappear on each per-suite page after the change.

### DIFF
--- a/Test/CLAUDE.md
+++ b/Test/CLAUDE.md
@@ -22,11 +22,19 @@ cd Test/Framework/TestWebApplication
 npx serve .
 
 # 3. Open http://localhost:3000/TestPage.htm in any browser
+#    (this is now an index page; pick a per-suite page from the list)
 ```
+
+> **Each test suite runs on its own page.** Co-loading multiple NScript bundles
+> on a single page corrupts `Function.prototype` runtime metadata because the
+> minifier picks per-IIFE letters that collide on the shared prototype (see
+> issue #51). `TestPage.htm` is now a link index; the actual suites are at
+> `Sunlight.Framework.TestPage.htm`, `Sunlight.Framework.UI.TestPage.htm`,
+> `DataTestPage.htm`, and `TodoApp.TestPage.htm`.
 
 ### How It Works
 
-C# test code using `[TestFixture]` / `[Test]` attributes from the `SunlightUnit` framework gets compiled to JavaScript by the NScript compiler. The `TestGenerator` plugin (`Sources/Compiler/NScript.Converter.Plugins/TestGenerator.cs`) scans for these attributes and emits `QUnit.module()` / `QUnit.test()` registration calls in the generated JS. The `TestPage.htm` page loads QUnit and the generated scripts, and QUnit runs everything automatically.
+C# test code using `[TestFixture]` / `[Test]` attributes from the `SunlightUnit` framework gets compiled to JavaScript by the NScript compiler. The `TestGenerator` plugin (`Sources/Compiler/NScript.Converter.Plugins/TestGenerator.cs`) scans for these attributes and emits `QUnit.module()` / `QUnit.test()` registration calls in the generated JS. Each suite-specific QUnit page (e.g. `Sunlight.Framework.TestPage.htm`) loads QUnit plus exactly one generated bundle, and QUnit runs everything automatically.
 
 ### Pipeline
 
@@ -37,7 +45,7 @@ C# [TestFixture]/[Test]  -->  dotnet build (Debug)  -->  nscript.exe + TestGener
                                                     GeneratedScripts/*.js
                                                               |
                                                               v
-                                                    TestPage.htm + QUnit 2.2.0
+                                                    Per-suite *.TestPage.htm + QUnit 2.2.0
                                                               |
                                                               v
                                                     Browser executes tests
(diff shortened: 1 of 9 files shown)


gautam-achieveai (commented):
[Gautam's review bot] **APPROVE**

Clean, well-documented avoidance fix for #51. The per-page isolation approach is correct - NScript bundles are designed as single-bundle-per-app, so test pages should mirror that.

**What's good:**
- New .htm pages mirror existing DataTestPage.htm structure exactly
- run-tests.js properly iterates suites, aggregates counts, and sets process.exitCode
- run-qunit.mjs already had this pattern and is correctly left untouched
- Documentation is thorough - the 'how to add a new suite' instructions now include the isolation constraint
- Troubleshooting section gains the key symptom (b.m is not a function -> co-loaded bundles)

**Minor observations (non-blocking):**
- The ~10-line HTML comment explaining per-IIFE collision is duplicated verbatim across all 4 test pages - when the root-cause fix lands, all need updating. Acceptable for static HTML.
- run-tests.js has no fail-fast option; all 4 pages always run even if one crashes. Fine for now.

No blocking issues. Ship it.

Issue #61, opened 2026-05-08
BondToAst deserializer: round-trip Roslyn-synthesised init/required/record-equality method-reference shapes


gautam-achieveai:
[Gautam's Dev bot]

## Context

Raised by reviewer on PR #60 (WI #47, Phase F3). When `Lang*Features.cs` fixtures are listed in the explicit `realScript` build list in `Test/Compiler/NScript.Csc.Lib.Test/TestResources.cs`, every method body in the fixture is parsed through the in-test pipeline:

```
DllBuilder.cs:175  topLevelBlock = toAst.ParseMethodBody(methodBody)
  -> BondToAst.ParseMethodBody(...)
     -> MemberReferenceDeserializer.DeserializeMethod(MethodSpecSer)
```

This path NREs on at least two classes of Roslyn-synthesised member-reference shape:

1. **Record equality / clone synthesis** (`Lang9RecordTests.cs`) — kept out of the explicit list since the records merge.
2. **`init` accessors and `required` member metadata** (`Lang11RequiredTests.cs`) — newly excluded in PR #60.

Both fixtures are still exercised end-to-end through NScript's own CSC because the MSBuild Framework build globs `Sources/Framework` and `Test/Framework`. The exclusion only avoids the in-test BondToAst pre-flight.

## Why this matters

The exclusion list is now two entries long and will grow with each new C# language wave that introduces synthesised member references (e.g. primary constructors, collection builder methods, ref-struct interfaces). Each excluded fixture is also a coverage gap on the Roslyn-driven CLR test path.

## What needs investigating

- Pinpoint the NRE site in `Sources/Compiler/JsCsc.Lib/MemberReferenceDeserializer.cs:DeserializeMethod(MethodSpecSer)` (likely a null `declaringType` flowing into `GetTypeNameMaps`, or a `Resolve()` returning null on a synthesised reference shape).
- Decide whether the fix is at the deserializer (more permissive resolution / better fallback) or at the serializer (record more shape data so the deserializer has what it needs).
- Once fixed, fold `Lang9RecordTests.cs` and `Lang11RequiredTests.cs` back into the explicit `realScript` file list and confirm the Test CLR step is green.

## Acceptance criteria

- `MemberReferenceDeserializer.DeserializeMethod(MethodSpecSer)` resolves Roslyn-synthesised `init` setters, `required` member references, and record `Equals`/`<Clone>$` shapes without NRE.
[...]

PR #75, opened 2026-05-13
[Gautam's Dev bot] WI-71: hoist switch-arm declaration-pattern locals (R2 of #71)

### LINKED ISSUE #71: Sub-issue: Runtime gaps - V8 execution snapshots + fixes for C# 9-13
## Summary
Seven C# 9-13 features bind cleanly and produce a compile-only fixture today
but raise runtime errors (or wrong output) when executed under V8. For each
gap, this sub-issue requires (1) an execution-validated `Lang*ExecutionTests`
fixture that reproduces the failure, (2) the fix in the appropriate pipeline
seam, (3) a snapshot under `Test/Compiler/NScript.Converter.Test/JsExecutionTests/Results/`,
and (4) a `[DataRow]` entry in `SimpleClassGenerationTests.TestMcs`.

Some of these gaps may warrant their own follow-up issues if the fix turns
out to be large; in that case file the spinoff and check the row here when
the spinoff lands.

## Scope

- [ ] **Record value-equality at runtime.** Roslyn's synthesised `Equals`
      calls `EqualityComparer<T>.Default`, which NScript's
      `Sources/Framework/mscorlib/Collections/Generic/EqualityComparer.cs`
      currently returns `null` from. Fix the facade so member-wise compare
      works for primitive + reference types; emit a snapshot from
      `Lang9RecordExecutionTests::RecordValueEquality`. **Blocked by #61**
      (BondToAst deserializer round-trip for the synthesised
      method-reference shapes).
- [ ] **Declaration patterns inside `switch` arms.** `o switch { string s
      => ..., int i => ..., }` raises `InvalidOperationException` from
      `BoundAstToAstBase.GetLocalVariable` because the switch-arm scope is
      not pushed onto `scopeBlockStack` before the arm's bound local is
      visited. Push the arm scope in
      `Sources/Compiler/NScript.Csc.Lib/BoundAstToAstBase.cs` and snapshot in
      `Lang9PatternExecutionTests`.
- [ ] **`ParenthesizedPattern` on boxed scrutinee.** `object o = 5; o is (5)`
      returns `false` through V8. Fix the lowering in `PatternMatcher.LowerToCondition`
      so the boxed-int / int-constant compare unwraps consistently with the
      non-parenthesised form, then snapshot under `Lang10ExecutionTests` (new
      class) or extend `Lang9PatternExecutionTests`.
- [ ] **`record struct` with `with` expression.** Roslyn does not synthesise
      `<Clone>$` for record structs; today `BondToAst.ParseWithExpression`
      raises a deliberate `NotImplementedException`. Route struct `with`
      through NScript's struct codegen path (which today does not preserve
      value-copy semantics on assignment — value-copy fix is a prerequisite).
      Snapshot in `Lang9RecordExecutionTests::RecordStructWith`.
- [ ] **Primary-ctor captured-parameter read from non-ctor method bodies.**
      `Counter.Reset()`'s `_count = initial;` reads the primary-ctor parameter
      `initial`, which `BondToAst.ParseArgumentVariable` cannot resolve.
      Lower captured-parameter reads from sibling method bodies onto the
      synthesised private backing field (mirror Roslyn's IL emission). Snapshot
      in `Lang12PrimaryCtorExecutionTests` (the existing F2 fixtures
      `PrimaryCtorOnClass` / `PrimaryCtorWithBaseCall` / `PrimaryCtorMultipleParams`
      can light up once the read path is wired).
- [ ] **C# 13 method-group natural type — overload-pruning runtime.**
      `Lang13Features.cs::MethodGroupOverloadPruning` raises
      `ConverterLocationException: Can't access generic type (!!0) if they
      are ignored` — the constraint-pruned generic candidate is DCE'd yet a
      downstream type reference still resolves to its `!!0` type parameter.
      Either preserve the candidate metadata during DCE or rewrite the type
      reference; snapshot in `Lang13ExecutionTests::MethodGroupOverloadPruning`.
- [ ] **`params IList<T>` / `ICollection<T>` / `IReadOnlyList<T>` /
      `IReadOnlyCollection<T>` member access.** `xs.Count` against an
      interface-typed receiver dispatches through the interface-suffixed
      getter slot (e.g. `V_get_Count_<suffix>`), which the synthesised
      `List<T>` instance does not expose — V8 raises `TypeError`. Fix the
      interface dispatch lowering or expose the suffixed slot on `List<T>`;
      snapshot in `Lang13ExecutionTests` and retire the compile-only
      `ParamsCollectionsInterfaceCount_RuntimeGap` placeholder.

## Acceptance
- [ ] Each row above ships a `Lang*ExecutionTests` entry under
      `Test/Compiler/NScript.Converter.Test/JsExecutionTests/` with a matching
      `Results/<TestName>.txt` snapshot and a `[DataRow(...)]` entry in
      `SimpleClassGenerationTests.TestMcs`.
- [ ] The matching `docs/language/csharp9-13-status.md` row is promoted to
      `✅ Supported` (or, where the fix proves out-of-scope, demoted to
      `❌ Out of scope` with a clear reason).
- [ ] No regression in `Lang9PatternExecutionTests` /
      `Lang13ExecutionTests` (the V8 snapshots that landed under #63).
- [ ] Each row may be closed individually as a spinoff issue if the fix
      proves large; in that case this issue tracks the rollup.

## Out of scope
- Cheap fixture promotions (Sub-issue A).
- BCL facade additions (Sub-issue C).
- Feature implementation for shapes that don't bind today (Sub-issue D).

## Parent
Sub-issue of #69.

## Related
- #61 — prerequisite for the record value-equality row.

gautam-achieveai:
[Gautam's Dev bot]

## Summary

Closes #71 (partial — R2 only; other rows R3/R5/R6/R7 will land in follow-up PRs per the plan's "keep each row independently reviewable" sequencing).

Fixes the R2 runtime gap from the approved plan in [comment #4438892908](https://github.com/achieveai/NScript/issues/71#issuecomment-4438892908): declaration patterns inside switch-expression arms.

### What was broken

```csharp
private static string Describe(object o) => o switch
{
    string s => "s:" + s,
    int i    => "i:" + i.ToString(),
    bool b   => "b:" + (b ? "t" : "f"),
    null     => "null",
    _        => "?"
};
```

threw `InvalidOperationException` from `BoundAstToAstBase.GetLocalVariable` at Stage 1 compile time. The arm-local's `ScopeDesignatorOpt` is the arm syntax itself, but switch expressions don't push their arms as block scopes (they lower to nested ternaries with no JS block to bind into) — so the direct designator match in `GetLocalVariable` never found a corresponding entry on `scopeBlockStack`.

### Fix shape

Single change in `Sources/Compiler/NScript.Csc.Lib/BoundAstToAstBase.cs::GetLocalVariable`: when the direct designator match fails, walk the designator's syntax ancestors until one matches an active scope. This hoists the local to the nearest enclosing block — its natural JS binding site for the inline `(local = AsType(T, scrutinee)) != null` test emitted by `PatternMatcher.LowerDeclarationPattern`. No serializer changes; no Stage 2 changes.

Arm-locals with the same C# name across multiple arms share the JS slot harmlessly: `VariableCollector.CreateVariable` auto-disambiguates colliding emitter names via `_0`/`_1` suffixes, and each `LocalSymbol` still resolves uniquely through `ResolveVariable`.

### Coverage

- New runtime test: `Lang9PatternExecutionTests.TestSwitchDeclarationArm` — exercises `string`, `int`, `bool`, and `null` arms against a boxed `object` scrutinee.
- Snapshot extended in `Lang9PatternExecutionTests.txt`.
- Existing `[DataRow(... "Lang9PatternExecutionTests", "Main" ...)]` in `SimpleClassGenerationTests.TestMcs` already covers the new helper via the updated `Main()`.

### Docs

`docs/language/csharp9-13-status.md`: the "Pattern matching — type pattern (without var)" row is promoted from ⚠️ Partial to ✅ Supported; the switch-arm `InvalidOperationException` caveat is replaced with a description of the hoist mechanism and the runtime-gating fixture.

### Why this PR is R2-only

The plan called out PR sequencing to keep each row independently reviewable. R3/R5/R6/R7 each touch different stages (lowering, codegen, serializer) and warrant separate review surfaces. R1 stays blocked on #61 (record value-equality via `EqualityComparer<T>.Default`); R4 (record-struct `with`) remains deferred.

## Test plan

- [x] Debug build: `dotnet build NScript_Full.sln -c Debug` → 0 errors
- [x] V8 snapshot run on Windows CI: `dotnet test Test/Compiler/NScript.Converter.Test/NScript.Converter.Test.csproj --filter "FullyQualifiedName~SimpleClassGenerationTests"` (Linux/WSL local environment can't run the V8 harness — `TestResources` uses Windows-style relative paths to compile the test mscorlib; verified the fixture compiles end-to-end at Stage 1 + Stage 2)
- [x] Verify the snapshot diff (only inserted lines 16-19: `s:hello`, `i:42`, `b:t`, `null`) matches V8 output on CI

---

_CI run [25807147931](https://github.com/achieveai/NScript/actions/runs/25807147931) — `Test Converter` step green on commit `03d7d6e`: the new `Lang9PatternExecutionTests.TestSwitchDeclarationArm` snapshot ran end-to-end through V8 and the inserted lines (`s:hello`, `i:42`, `b:t`, `null`) match the expected output. Both Test-plan checkboxes ticked._


gautam-achieveai:
[Gautam's review bot] **APPROVE** ✅

## Summary

Clean, well-scoped fix for the `InvalidOperationException` crash when compiling declaration patterns inside switch-expression arms. The approach (walking syntax ancestors to find the nearest enclosing block scope) is correct and minimal.

## What I verified

1. **Correctness of the ancestor walk**: Reference equality on syntax nodes is the Roslyn convention. Null-safety handled via `?.Parent`. The existing `InvalidOperationException` remains as the final safety net if no ancestor matches.

2. **No name collision risk**: `VariableCollector.CreateVariable` auto-disambiguates via `_N` suffix when multiple arm-locals (e.g. `s`, `i`, `b`) are hoisted to the same enclosing block.

3. **Performance**: The fallback is O(depth x stack_size) but only triggers for the edge case (pattern locals in switch arms). Both depth and stack size are small (2-5 typically). Compile-time only.

4. **Test coverage**: 4 arm types exercised (`string`, `int`, `bool`, `null`) plus wildcard discard. Snapshot ordering consistent with `Main()` call sequence.

5. **Docs**: Status promoted from Partial to Supported with accurate explanation.

## Minor observations (non-blocking)

- The fix also implicitly handles `var` patterns inside switch arms (same `ScopeDesignatorOpt` mechanism) — worth a future test.
- Nested switch expressions would also be fixed by the ancestor walk — not tested but the approach is generic.

## Context

- Linked to Bug #71 (Runtime gaps - V8 execution snapshots + fixes for C# 9-13)
- This is R2 in a planned sequence of independent PRs (R1 blocked on #61; R3/R5/R6/R7 pending as follow-ups)

gautam-achieveai:
I don't see that you've completed 2 remaining task description

gautam-achieveai:
[Gautam's Dev bot]

Apologies for leaving those two test-plan boxes unchecked — they were actually already satisfied by CI on the merge candidate. I've now ticked them and added a confirmation note to the PR description.

What the CI run verifies:

- **V8 snapshot run on Windows CI**: Action run [25807147931](https://github.com/achieveai/NScript/actions/runs/25807147931), step `Test Converter` on commit `03d7d6e` — green. That step runs `Test/Compiler/NScript.Converter.Test/NScript.Converter.Test.csproj`, which includes `SimpleClassGenerationTests.TestMcs` with the `[DataRow(... "Lang9PatternExecutionTests", "Main" ...)]` entry. That row executes the new `TestSwitchDeclarationArm` helper end-to-end through the V8 harness.
- **Snapshot diff matches V8 output**: same step is the snapshot consumer — it diffs the V8 stdout against `Lang9PatternExecutionTests.txt`. Passing = the four inserted lines (`s:hello`, `i:42`, `b:t`, `null`) match V8 exactly. A drift would have failed the step.

I couldn't tick them on PR open because the local Linux/WSL environment can't run the V8 harness (`TestResources` uses Windows-style relative paths to compile the test mscorlib), so confirmation had to wait until Windows CI completed — which it has.

PR #78, opened 2026-05-15
[Gautam's Dev bot] WI-73: target-typed new() + module initializers (Child 1 + 2)

gautam-achieveai:
[Gautam's Dev bot]

## Summary

Implements two of the four deferred C# 9 features tracked under #73, in the smallest-blast-radius order from the approved plan (comment #4455727534).

**Child 1 — Target-typed `new()` (C# 9)**

- Add `ConversionKind.ObjectCreation` to `VisitConversion`'s pass-through cluster (alongside `Deconstruction`/`SwitchExpression`/`CollectionExpression`). Forwards to the inner `BoundObjectCreationExpression` so the ordinary constructor pipeline emits the constructed instance unchanged.
- Compile-only fixtures in `Lang9Features.cs::TargetTypedNew*`: parameterless, with-initializer, with-args, and as-return shapes.
- V8 execution snapshot in new `Lang9FeatureExecutionTests` covers each shape with field/ctor assertions.

**Child 2 — Module initializers (C# 9)**

- New `System.Runtime.CompilerServices.ModuleInitializerAttribute` facade in `mscorlib`, paralleling `EntryPointAttribute` (sealed, `[NonScriptable]`, `AttributeTargets.Method`).
- `ConverterKnownReferences.ModuleInitializerAttribute` property looks up the BCL symbol for Stage-2 attribute scanning.
- `Builder.GetModuleInitializers` collects every static / parameterless / void method carrying `[ModuleInitializer]` in stable declaration order (Mono.Cecil `module.Types` enumerates in metadata-token order — close-enough proxy for the CLR's "metadata token order" guarantee in a single-bundle world).
- `Builder` emits invocation statements for each initializer **before** the entry-point call, matching the entry-point emission shape (same `ResolveFunctionName` + `JST.IdentifierExpression` plumbing, so minifier rename + dead-code elimination apply uniformly).
- V8 execution snapshot covers (a) initializer ran before `Main` and (b) ordering across two tracker types is deterministic.

## Scope

This PR closes **Children 1 + 2** of the rollup #73. Children 3 (extended property patterns `{ A.B: ... }`) and 4 (list patterns `[1, 2, ..]`) remain open follow-ups — each requires proto-tag reservations and `PatternMatcher` visitor work and is split into its own PR per the plan's smallest-blast-radius ordering. The rollup stays open until Children 3 + 4 land.

Relates to #73

## Test plan

- [ ] CI green on `dotnet build NScript_Full.sln -c Debug` (Linux sandbox hit disk-full; relying on Windows CI for full-solution validation)
- [ ] `SimpleClassGenerationTests.TestMcs(Lang9FeatureExecutionTests, Main, ...)` snapshot matches `Lang9FeatureExecutionTests.txt`:
  - `ttn:par:1`, `ttn:init:11`, `ttn:args:42`, `ttn:ret:7` (target-typed new in 4 shapes)
  - `mi:ran:99` (module initializer ran before Main)
  - `mi:order:AB` (deterministic ordering across types)
- [ ] All other compiler test suites remain green


gautam-achieveai (commented):
## [Gautam's review bot] Deep Review — PR #78

### Context
This PR implements two of four deferred C# 9 features tracked under #73:
- **Child 1:** Target-typed `new()` — adds `ConversionKind.ObjectCreation` to VisitConversion's pass-through cluster
- **Child 2:** Module initializers — scans for `[ModuleInitializer]` methods and emits calls before entry point

### Verdict: REQUEST CHANGES

The production code changes are architecturally sound, but the test coverage is **broken** — all execution tests for this PR are dead code that will never successfully run.

---

### :red_circle: BLOCKER: `Lang9FeatureExecutionTests.cs` missing from `TestResources.cs`

The new test file is registered in `SimpleClassGenerationTests.cs` DataRow (line 61) but is **NOT** listed in `TestResources.cs` (neither the release list at lines 375-414 nor the debug list at lines 440-481).

Without this listing, `TestAssemblyLoader.GetMethodDefinition("Lang9FeatureExecutionTests", "Main", ...)` will throw — the class doesn't exist in the compiled test assembly. Every other `*ExecutionTests.cs` file (Lang8, Lang9Pattern, Lang13) is properly registered. This one was missed.

**Fix:** Add `@"Lang9FeatureExecutionTests.cs"` after line 382 and after line 448 in `TestResources.cs`.

---

### :red_circle: HIGH: Module initializer tests cannot pass — test harness doesn't exercise Builder pipeline

Even with the TestResources fix above, the module initializer assertions will fail:

- `JsExecutionHelper.GetScript()` only converts a single entry-point method via `runtimeScopeManager.Convert()` and emits a call to it
- It does **NOT** call `GetModuleInitializers()` or emit initializer invocations
- Therefore `ModuleInitTrackerA.Init()` / `ModuleInitTrackerB.Init()` will **never execute**
- Expected output `mi:ran:99` and `mi:order:AB` will fail against actual `mi:ran:0` and `mi:order:`

The module initializer emission logic in `Builder.cs` (lines 227-235 + 461-496) has **zero effective test coverage**.

**Suggested fixes (pick one):**
1. **(Preferred)** Split the test — keep target-typed new tests in `Lang9FeatureExecutionTests` (they work via demand-driven conversion), and create a separate integration test that exercises `Builder.Generate()` for module initializers
2. **(Alternative)** Modify `JsExecutionHelper` to optionally discover and emit module initializer calls for the test assembly

---

### :large_orange_diamond: MEDIUM: `TryGetModuleDefinition` return value not checked

`Builder.cs:466` — the bool return from `TryGetModuleDefinition` is discarded. If the module lookup fails, the next line (`foreach (var item in module.Types)`) throws a NullReferenceException caught by the outer `catch(Exception)` — producing the unhelpful message *Object reference not set to an instance of an object* instead of *assembly not found*.

This copies the pre-existing pattern from `GetEntryPoint` (line 421), and every other caller in the codebase checks the return value properly (`ClrKnownReferences.cs:972`, `MemberReferenceDeserializer.cs:643`, etc.).

**Suggested fix:**
`csharp
if (!context.ClrContext.TryGetModuleDefinition(mainAssembly, out module))
    return new List<MethodDefinition>();
`

---

### :white_circle: LOW: Nested types not scanned for module initializers

`ModuleDefinition.Types` (Mono.Cecil) returns only top-level types. Module initializers in nested types would be silently ignored. The CLR spec does allow `[ModuleInitializer]` on nested type methods. Same limitation exists in `GetEntryPoint` — pre-existing pattern. Not a blocker, but worth a comment or follow-up.

---

- Target-typed `new()` implementation is **correct and minimal** — single pass-through case
- Module initializer emission logic is architecturally sound (correct layer, correct ordering)
- `ModuleInitializerAttribute` facade follows established conventions
- Documentation updates in `csharp9-13-status.md` are thorough and accurate
- The *smallest-blast-radius* PR splitting strategy is good engineering practice

### Agents Used
architecture-review, performance-review, exception-handling-review, test-coverage-review

Test/Compiler/NScript.Converter.Test/JsExecutionTests/SimpleClassGenerationTests.cs
```diff
@@ -58,6 +58,7 @@ public void Setup()
         [DataRow(TestType.Debug, "TestAsyncAwait", "Main", false, true, true)]
         [DataRow(TestType.Debug, "Lang8FeatureExecutionTests", "Main", false, true, false)]
         [DataRow(TestType.Debug, "Lang9PatternExecutionTests", "Main", false, false, false)]
+        [DataRow(TestType.Debug, "Lang9FeatureExecutionTests", "Main", false, false, false)]
```
gautam-achieveai, line 61: [Gautam's review bot] :red_circle: **BLOCKER**: This DataRow references `Lang9FeatureExecutionTests` but the file is NOT listed in `TestResources.cs` (neither the release source list at lines 375-414 nor the debug list at lines 440-481). `TestAssemblyLoader.GetMethodDefinition` will fail because the class won't be compiled into the test assembly.

**Fix:** Add `@"Lang9FeatureExecutionTests.cs"` to both source arrays in `TestResources.cs` (after `Lang9PatternExecutionTests.cs` at line 382 and line 448).

Test/Framework/RealScript/Lang9FeatureExecutionTests.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        private static Holder Make() => new(7);
+
+        // `[ModuleInitializer]` static methods on `ModuleInitTrackerA` /
+        // `ModuleInitTrackerB` are emitted before the entry point by
+        // `Builder.cs`. The first assertion verifies the call ran; the
+        // second verifies deterministic ordering across types — both
+        // initializers append to the same buffer in declared order.
+        private static void TestModuleInitializerRanBeforeMain()
+        {
```
gautam-achieveai, line 69: [Gautam's review bot] :red_circle: **HIGH**: These module initializer test methods (`TestModuleInitializerRanBeforeMain` and `TestModuleInitializerOrderingAcrossTypes`) **cannot pass** with the current test harness.

`JsExecutionHelper.GetScript()` only converts and calls the single entry-point method via `runtimeScopeManager.Convert()`. It does NOT discover `[ModuleInitializer]` methods or emit their invocations — that logic lives in `Builder.cs` which the test harness doesn't use.

Result: `ModuleInitTrackerA.Init()` and `ModuleInitTrackerB.Init()` will never execute. `Marker` stays 0 (not 99), `Trace` stays empty (not "AB").

**Suggested fix:** Either split these into a separate integration test that exercises `Builder.Generate()` end-to-end, or modify `JsExecutionHelper` to optionally emit module initializer calls.

Sources/Compiler/NScript.Converter/Builder.cs
```diff
@@ -434,5 +451,48 @@ private MethodDefinition GetEntryPoint(ConverterContext context, string mainAsse
...
+        /// carries <c>[ModuleInitializer]</c>. Returns them in stable declaration order
+        /// (type metadata-token order, then method declaration order) — close-enough
+        /// proxy for CLR's "metadata token order" guarantee in a single-bundle world.
+        /// </summary>
+        private List<MethodDefinition> GetModuleInitializers(ConverterContext context, string mainAssembly)
+        {
+            var result = new List<MethodDefinition>();
+
+            ModuleDefinition module;
+            context.ClrContext.TryGetModuleDefinition(mainAssembly, out module);
```
gautam-achieveai: [Gautam's review bot] :large_orange_diamond: **MEDIUM**: The `bool` return from `TryGetModuleDefinition` is discarded. If module lookup fails, `module` is null and the `foreach` on the next line throws `NullReferenceException`.

Every other caller in the codebase checks this return value (`ClrKnownReferences.cs:972`, `MemberReferenceDeserializer.cs:643,670`, `MemberReferenceConverter.cs:49,63`). The pre-existing `GetEntryPoint` (line 421) has the same gap, but that's not a reason to replicate it.

Suggested fix:
```csharp
if (!context.ClrContext.TryGetModuleDefinition(mainAssembly, out module))
    return new List<MethodDefinition>();
```

gautam-achieveai:
tests failing?

Test/Compiler/NScript.Converter.Test/JsExecutionTests/SimpleClassGenerationTests.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 3f2bcae5 — added `@"Lang9FeatureExecutionTests.cs"` to both `realScript` (line 382) and `realScript.debug` (line 449) source lists in `TestResources.cs`. Also registered the new `Runtime\CompilerServices\ModuleInitializerAttribute.cs` facade in the mscorlib list (line 204) so the in-test MCS compilation can resolve the attribute reference.

Test/Framework/RealScript/Lang9FeatureExecutionTests.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 3f2bcae5 — went with option 2 (wire `JsExecutionHelper` to discover and emit module initializer calls). The helper now scans `methodDefinition.Module` for `[ModuleInitializer]`-tagged static parameterless void methods (mirroring `Builder.GetModuleInitializers`), adds them to the `Convert()` input list so the runtime pulls in their bodies, and emits explicit JST `ExpressionStatement` calls in declaration order **before** the entry-point call. This matches `Builder.Generate`'s pre-Main ordering and keeps the V8 snapshot fixture authoritative for both target-typed `new()` and module initializer side effects. Discovery is contained in a small private helper (`CollectModuleInitializers`) — Builder remains the production source-of-truth.

Sources/Compiler/NScript.Converter/Builder.cs
gautam-achieveai: [Gautam's Dev bot]

Fixed in 3f2bcae5 — `GetModuleInitializers` now early-returns an empty list when `TryGetModuleDefinition` fails, so a missing main assembly surfaces as `no initializers found` rather than a downstream NRE. Left `GetEntryPoint` (line 421) untouched — agree the pre-existing pattern is buggy too, but that's outside this PR's blast radius and noted as a follow-up.

gautam-achieveai (commented):
[Gautam's review bot] **Incremental Review - Iteration 2** (commit 3f2bcae5)

## Previous Findings - All Addressed ✅

| # | Severity | Issue | Status |
|---|----------|-------|--------|
| 1 | BLOCKER | Lang9FeatureExecutionTests.cs missing from TestResources.cs | ✅ Fixed |
| 2 | HIGH | Module initializer tests dead (harness doesn't emit initializer calls) | ✅ Fixed |
| 3 | MEDIUM | TryGetModuleDefinition return value discarded - potential NRE | ✅ Fixed |

## Verification of Fixes

- **Builder.cs**: Clean defensive early-return. ✅
- **JsExecutionHelper.cs**: `CollectModuleInitializers` correctly mirrors production filters (interface/generic type skip, static + parameterless + void + attribute checks). Call emission ordering (initializers before entry-point) matches `Builder.Generate`. ✅
- **TestResources.cs**: Entries in correct alphabetical position in both release/debug lists. ✅

## Minor Observation (non-blocking)

The `CollectModuleInitializers` helper in JsExecutionHelper is a structural duplicate of `Builder.GetModuleInitializers`. If production logic evolves (e.g., nested type scanning per CLR spec), the test helper could silently drift. Consider extracting a shared static utility long-term — but this is fine for now given the test harness's intentional unit-level scope.

## Re: 'tests failing?' thread

The CI failures were caused by the BLOCKER (missing test file registration). The iteration 2 commit should resolve this — all test infrastructure is now properly wired.

**Verdict: APPROVE** — All blocking issues resolved. Implementation is correct and well-structured.

gautam-achieveai:
[Gautam's Dev bot] Pushed fixup `fc530a27` to address the iteration-2 CI failure.

**Root cause:** the V8 snapshot reported `ttn:par:[object Object]` for every int operand. NScript boxes primitive `int` through `System.Object` for `string + int` concat — `String.Concat(string, object)` calls `obj.ToString()` on a wrapper that doesn't override `toString()` under the V8 host, so it prints `[object Object]` rather than the numeric value.

**Fix:** call `.ToString()` explicitly on int operands in `Lang9FeatureExecutionTests.cs`, mirroring the existing safe pattern at `Lang13Features.Take` (line 56). The string-typed operand (`ModuleInitTrackerB.Trace`) already worked — `mi:order:AB` was produced correctly in iteration 2, confirming the module-initializer wiring landed in `3f2bcae5` is sound; the only residual failure was int formatting.

Awaiting Windows CI re-run on `fc530a27`.

gautam-achieveai:
[Gautam's review bot] **Iteration 3 incremental review** (fc530a27 vs 3f2bcae5)

:white_check_mark: **APPROVE** — The `.ToString()` fix is correct and consistent across all 5 int-operand sites. The explanatory comment about NScript's int boxing under V8 is valuable context for future maintainers. Line 75 (`ModuleInitTrackerB.Trace`) correctly left unchanged since it's already `string`-typed.

All 3 original findings (BLOCKER + HIGH + MEDIUM) from iteration 1 remain fixed. No new issues introduced.

---
**Review path:** Incremental (iteration 3 only — 1 file, 9 insertions / 5 deletions)
**Agents used:** Manual review (change too small for agent dispatch)
**Verdict:** APPROVE — ready to merge pending CI green.

gautam-achieveai:
[Gautam's Dev bot] Pushed fixup `9f4ade93` to address the iteration-3 CI failure.

**Root cause:** the V8 script produced the exact expected output (`ttn:par:1` / `ttn:init:11` / `ttn:args:42` / `ttn:ret:7` / `mi:ran:99` / `mi:order:AB`) — confirming `fc530a27` fixed the int-formatting issue — but the snapshot file `Lang9FeatureExecutionTests.txt` was not declared as `<EmbeddedResource>` in `NScript.Converter.Test.csproj`. `GetResourceString` returned empty, so the assert compared the actual console output against the raw `output` parameter ("Lang9FeatureExecutionTests") and failed.

**Fix:** add the two missing csproj entries (mirroring the existing `Lang9PatternExecutionTests` lines):
```xml
<None Remove="JsExecutionTests\Results\Lang9FeatureExecutionTests.txt" />
<EmbeddedResource Include="JsExecutionTests\Results\Lang9FeatureExecutionTests.txt" />
```

Awaiting Windows CI re-run on `9f4ade93`.

gautam-achieveai:
[Gautam's review bot] **Iteration 4 incremental review** (9f4ade93 vs fc530a27)

:white_check_mark: **APPROVE** - The `<EmbeddedResource>` registration fix is correct. The `.txt` snapshot file exists on disk with the expected content (`ttn:par:1` / `ttn:init:11` / `ttn:args:42` / `ttn:ret:7` / `mi:ran:99` / `mi:order:AB`), and the two new csproj entries mirror the existing `Lang9PatternExecutionTests` pattern exactly.

All 3 original findings (BLOCKER + HIGH + MEDIUM) from iteration 1 remain addressed. No new issues introduced across iterations 2-4.

---
**Review path:** Incremental (iteration 4 only - 1 file, 2 insertions)
**Agents used:** Manual review (2-line csproj fix, no agent dispatch needed)
**Verdict:** APPROVE - ready to merge pending CI green.

**Note:** This is the 3rd consecutive CI-only fixup (test registration -> int formatting -> embedded resource). Consider running the full test suite locally before pushing next time to catch these build/resource wiring issues earlier.

PR #81, opened 2026-05-17
[Gautam's Dev bot] WI-79: fix Exception subclass prototype seed for [Extended] base types

### LINKED ISSUE #79: Exception subclass prototype uses `new Error()` — crashes at module init when base is native Extended type
## Summary

Custom exception classes (e.g. `NotImplementedException`) crash at module initialization.
`System.Exception` is mapped to native `Error` via `[ScriptName("Error"), Extended]`. When
NScript generates prototype setup for a derived type, it emits `new Error()`, which can
throw before the prototype chain is established.

## Steps to Reproduce

1. Define a class inheriting from `System.Exception` in C#
2. Use it in code (e.g. `throw new NotImplementedException()`)
3. Compile with NScript
4. Load the generated JS module — it throws during initialization

## Actual Behavior

Generated module crashes during type initialization:

```javascript
NotImplementedException_Q.defaultConstructor_d = NotImplementedException_factory_cS;
ptyp__a = new Error();   // ← throws here (line 2485:11 in AgentWorkBench.js)
NotImplementedException_Q.prototype = ptyp__a;
```

Stack trace observed:
```
Error
    at AgentWorkBench.js:2485:11
    at AgentWorkBench.js:6268:3
```

## Expected Behavior

Module loads successfully; custom exception types can be thrown and caught at runtime.

## Root Cause

`GetBasePrototypeExpression()` in `TypeConverter.cs:1605` checks
`!this.Context.IsExtended(this.typeDefinition)` — whether the **current** type is
Extended. For `NotImplementedException` (not Extended), this check passes and the method
emits `new Error()` to set up the prototype chain.

But `Error` is itself a native Extended type (`[ScriptName("Error"), Extended]`), and
calling `new Error()` to construct a prototype object can throw before the chain is
established.

The guard must also check whether the **base type** is Extended.

## Fix

Use `Object.create(BaseType.prototype)` instead of `new BaseType()` when the base is an
Extended (native) type. This sets up the correct prototype chain without invoking the
native constructor.

After fix, generated code:

```javascript
ptyp__a = Object.create(Error.prototype);   // safe — no constructor call
NotImplementedException_Q.prototype = ptyp__a;
```

**Changed file:** `Sources/Compiler/NScript.Converter/TypeSystemConverter/TypeConverter.cs`
— `GetBasePrototypeExpression()` — adds `IsExtended(BaseType)` guard and emits
`Object.create(BaseType.prototype)` for Extended base types.

All 410 `NScript.Converter.Test` tests pass after the fix.

gautam-achieveai:
[Gautam's Dev bot]

## Summary
Fixes #79 — C# subclasses of `System.Exception` (and any user type whose base is an `[Extended]` facade) crashed at JS module init.

`GetBasePrototypeExpression` in `TypeConverter.cs` unconditionally emitted `ptyp_ = new BaseType()` to seed the prototype chain. When the base is an `[Extended]` facade like `System.Exception` → native `Error`, that runs the native constructor during module init — before the prototype chain is wired — and the bundle fails to load.

Fix: when the base type is `[Extended]`, emit `Object.create(BaseType.prototype)` instead. The non-Extended path is unchanged.

## Changes
- `Sources/Compiler/NScript.Converter/TypeSystemConverter/TypeConverter.cs` — Extended-base branch in `GetBasePrototypeExpression`.
- `Test/Framework/RealScript/ExtendedBaseTypes.cs` — fixture: subclass `Exception`, instantiate from `Main`, print `Message`.
- `Test/Compiler/NScript.Csc.Lib.Test/TestResources.cs` — wire fixture into both `realScript` and `realScript.debug` file lists.
- `Test/Compiler/NScript.Converter.Test/JsExecutionTests/Results/ExtendedBaseExceptionExecutionTests.txt` — snapshot baseline (`boom`).
- `Test/Compiler/NScript.Converter.Test/NScript.Converter.Test.csproj` — register snapshot as embedded resource.
- `Test/Compiler/NScript.Converter.Test/JsExecutionTests/SimpleClassGenerationTests.cs` — `[DataRow]` for the new fixture.

## Test plan
- [ ] CI: full Release build green
- [ ] CI: `NScript.Converter.Test` `SimpleClassGenerationTests.TestMcs` `ExtendedBaseExceptionExecutionTests` row passes (V8 executes, snapshot matches `boom`)
- [ ] No drift in other JS snapshots (Extended-base case only fires for user types whose direct base is an `[Extended]` facade)

Closes #79


gautam-achieveai (commented):
[Gautam's review bot] **Deep Review — PR !81 / Bug #79**

## Summary

The fix logic in `GetBasePrototypeExpression()` is **correct in concept** — using `Object.create(BaseType.prototype)` avoids invoking the native constructor at module init. However, the test added to validate it is **failing**, which blocks merge.

## Findings

### CRITICAL: New test `ExtendedBaseExceptionExecutionTests` fails

```
Assert.AreEqual failed. Expected:<boom>. Actual:<>.
```

Running the full `SimpleClassGenerationTests.TestMcs` suite: **356 pass, 2 fail** — the new `ExtendedBaseExceptionExecutionTests` row and a pre-existing `TestAsyncAwait` failure unrelated to this PR.

The empty output (not a V8 exception) indicates the module **loads successfully** (so the prototype seed crash is fixed), but `Console.WriteLine(ex.Message)` produces nothing. This suggests the `Message` property isn't propagating through the constructor chain when the base is a native `Error`. The prototype seed is correct, but the instance-level constructor call (`super()` / `Error.call(this, msg)`) may not be wiring `message` into the native Error correctly.

**Action needed:** Investigate why the generated JS constructor for `ExtendedBaseException` doesn't set the native `Error.message` on the instance. The test must pass before merge.

### LOW: `ScriptLiteralExpression` usage for `Object.create`

This is the **only** usage of `ScriptLiteralExpression` in the entire `NScript.Converter` project. The rationale is sound (native JS global shouldn't be minified), but it introduces a pattern that has no precedent in this codebase layer. Consider adding `Object.create` as a known reference if this pattern may recur for other native APIs.

## What looks good

- Outer null/guard checks properly protect the new branch
- `PrototypeField` and `BaseType` are resolved through the scope system (minification-safe)
- `IsExtended(TypeDefinition)` handles null from `Resolve()` gracefully
- No performance concern — single cached lookup per qualifying type at compile time
- Test fixture design is appropriate (subclass Exception, instantiate from Main, print Message)
- Test resource registration in both `TestResources.cs` arrays and `.csproj` is correct

## Verdict: REQUEST CHANGES

The compiler fix is architecturally sound but the validation test fails. The PR's own test plan requires the snapshot to match `boom`. Please fix the constructor chain propagation issue and ensure the test passes green.

Sources/Compiler/NScript.Converter/TypeSystemConverter/TypeConverter.cs
```diff
@@ -1610,6 +1610,35 @@ private Expression GetBasePrototypeExpression()
...
+                // to thread the prototype chain without invoking the native constructor.
+                if (this.Context.IsExtended(this.TypeDefinition.BaseType))
+                {
+                    return new MethodCallExpression(
+                        null,
+                        this.Scope,
+                        new ScriptLiteralExpression(
+                            null,
+                            this.Scope,
+                            "Object.create"),
```
gautam-achieveai, line 1626: [Gautam's review bot] This is the only usage of ScriptLiteralExpression in the entire NScript.Converter project. While correct (native JS global shouldn't be minified), consider whether Object.create should be added to ConverterKnownReferences as a well-known method if this pattern may recur for other Extended base types. Not blocking, just a design consideration.

Test/Framework/RealScript/ExtendedBaseTypes.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// subclass of <see cref="Exception"/> from <c>Main()</c>. Without the fix,
+    /// the bundle fails to load and <c>Main()</c> never runs.
+    /// </summary>
+    public class ExtendedBaseExceptionExecutionTests
+    {
+        public static void Main()
+        {
+            ExtendedBaseException ex = new ExtendedBaseException("boom");
+            Console.WriteLine(ex.Message);
+        }
```
gautam-achieveai, line 29: [Gautam's review bot] **Test is failing.** Expected output: boom, Actual: empty string.

The module loads successfully (crash is fixed), but ex.Message is empty. This suggests the generated JS constructor for ExtendedBaseException doesn't properly propagate the message argument to the native Error base.

In native JS, Error.message is set by the Error() constructor call, not inherited via prototype. So even though the prototype seed is now correct, the instance constructor needs to invoke something like Error.call(this, message) or set this.message explicitly.

Please investigate the generated constructor chain for this class and verify the message parameter reaches the native Error internals.

PR #84, opened 2026-05-17
Fix static-field cctor crash: emit all _tri_b() before concrete cctors

### LINKED ISSUE #83: Static field initializer using ObservableCollection<T> crashes: _tri_b() runs after cctor
## Summary

When a class has a `static readonly ObservableCollection<T>` field (where `T` is a user-defined type), the app crashes at startup with:

```
TypeError: Cannot read properties of undefined (reading 'defaultConstructor_c')
    at ObservableCollection$1____ctor_md
    at defaultConstructor_c
    at <ClassName>____cctor_ew
```

## Repro

```csharp
public class WorkspaceListViewModel : ObservableObject
{
    // Any static field that constructs ObservableCollection<UserType>
    internal static readonly ObservableCollection<TagChip> EmptyTagChips = new ObservableCollection<TagChip>();
}
```

## Root cause

NScript's generic instantiation uses a three-phase init pattern. The concrete generic class (`ObservableCollection_E(TagChip_M)`) is created in phase 1, but the internal `List<T>` backing store (`List$1_$T$__f`) is only assigned inside the `_tri_b()` callback in phase 3.

The generated module-level initialization code emits the static constructor call **before** `_tri_b()` runs:

```js
// phase 1: generic class created, but List<TagChip> inside closure is still undefined
ObservableCollection_$TagChip$__cf = ObservableCollection_E(TagChip_M);

// cctor runs here — calls new ObservableCollection<TagChip>() → CRASH
// because List$1_$T$__f is undefined
WorkspaceListViewModel____cctor_ew();

// phase 3: too late — this would have set List$1_$T$__f
ObservableCollection_$TagChip$__cf._tri_b();
```

Inside `ObservableCollection$1____ctor_md`:
```js
this.items_e = List$1_$T$__f.defaultConstructor_c(); // List$1_$T$__f is undefined → crash
```

`List$1_$T$__f` is a closure variable that only gets assigned when `_tri_b()` runs:
```js
ObservableCollection$1_$T$__b._tri_b = function() {
    if (__initTracker_i) return;
    __initTracker_i = true;
    List$1_$T$__f = List_x(T_d, true); // ← never reached before cctor
    ...
};
```

## Expected behavior

`_tri_b()` for all generic types used in a static field initializer should be called **before** the class's `cctor`. The codegen ordering should guarantee: instantiate generic → call `_tri_b()` → call `cctor`.

## Workaround

Avoid static field initializers that construct `ObservableCollection<T>`. Use lazy initialization instead:

```csharp
private static ObservableCollection<TagChip> _emptyTagChips;
internal static ObservableCollection<TagChip> EmptyTagChips =>
    _emptyTagChips ??= new ObservableCollection<TagChip>();
```

By the time any instance method or property is called, all `_tri_b()` calls have already completed, so this is safe.

## Related

This is related to #82 (tree-shaking + zero-arg getter stripping). Both are initialization-ordering issues in the NScript codegen for generic types.

## Environment

- Cs2Jsc: 1.1.3
- Sdk + framework: 1.1.1

gautam-achieveai:
## Summary

- Fixes #83: `static readonly ObservableCollection<UserType>` crash at startup
- Splits the module-init loop in `RuntimeScopeManager.Convert()` into two passes: all `_tri_b()` calls first, then all cctors
- Adds a `HasTypeRefInit` property to `TypeConverter` to distinguish generic types (with `_tri_b`) from concrete types

## Root Cause

Generic closure variables (e.g. `List$1_$T$__f` inside `ObservableCollection<T>`) are only initialized when `_tri_b()` runs. The old single-loop ordering placed `WorkspaceListViewModel.cctor()` before `ObservableCollection<TagChip>._tri_b()` whenever the generic type was added to `typesInitializedInOrder` by the third loop (not by the dependency callback), causing a crash on `List$1_$T$__f.defaultConstructor_c()`.

## Fix

Pass 1: for every type in `typesInitializedInOrder` where `HasTypeRefInit == true`, emit the `_tri_b()` call.  
Pass 2: for every type where `HasTypeRefInit == false`, emit the static constructor call.

## Test

Added `StaticCollectionHolder` + `StaticCollectionItem` classes and a regression test `TestStaticReadonlyObservableCollectionOfUserType` in `ObservableCollectionTests`. The test accesses a `static readonly ObservableCollection<StaticCollectionItem>` field — if the ordering bug is present, the module crashes before any test runs.

### DIFF
--- a/Sources/Compiler/NScript.Converter/TypeSystemConverter/RuntimeScopeManager.cs
+++ b/Sources/Compiler/NScript.Converter/TypeSystemConverter/RuntimeScopeManager.cs
@@ -530,7 +530,11 @@ namespace NScript.Converter.TypeSystemConverter
                     if (!converter.HasTypeRefInit)
                     { continue; }
 
-                    returnValue.Add(converter.InitializeTypeStatement(typeReference));
+                    var triB = converter.InitializeTypeStatement(typeReference);
+                    if (triB != null)
+                    {
+                        returnValue.Add(triB);
+                    }
                 }
 
                 // Pass 2: run static constructors for concrete (non-generic) types.
--- a/Sources/Compiler/NScript.Converter/TypeSystemConverter/RuntimeScopeManager.cs
+++ b/Sources/Compiler/NScript.Converter/TypeSystemConverter/RuntimeScopeManager.cs
@@ -518,20 +518,35 @@ namespace NScript.Converter.TypeSystemConverter
                                 Scope)));
                 }
 
-                // here we need to initialize all the static constructor for
-                // all these classes.
+                // Pass 1: call _tri_b() on all generic types before any concrete cctor runs.
+                // A concrete cctor may call new ObservableCollection<T>() which requires T's
+                // _tri_b() to have already set the closure variable for List<T>.
                 foreach (var typeReference in typesInitializedInOrder)
                 {
                     if (!typesDefinitionsUsed.ContainsKey(typeReference.Resolve()))
                     { continue; }
 
-                    var staticConstructor =
-                        typesDefinitionsUsed[typeReference.Resolve()]
-                            .InitializeTypeStatement(typeReference);
+                    var converter = typesDefinitionsUsed[typeReference.Resolve()];
+                    if (!converter.HasTypeRefInit)
+                    { continue; }
 
-                    if (staticConstructor != null)
+                    returnValue.Add(converter.InitializeTypeStatement(typeReference));
+                }
+
+                // Pass 2: run static constructors for concrete (non-generic) types.
+                foreach (var typeReference in typesInitializedInOrder)
+                {
+                    if (!typesDefinitionsUsed.ContainsKey(typeReference.Resolve()))
+                    { continue; }
+
+                    var converter = typesDefinitionsUsed[typeReference.Resolve()];
+                    if (converter.HasTypeRefInit)
+                    { continue; }
+
+                    var cctorCall = converter.InitializeTypeStatement(typeReference);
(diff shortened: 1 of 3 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** ✅ APPROVE

## Summary

Clean, well-targeted fix for #83. The two-pass split correctly ensures all `_tri_b()` calls (which set up generic closure variables) execute before any concrete cctor that might depend on them.

### What I verified:
- `InitializeTypeStatement()` returns either `_tri_b()` call OR cctor — never both. For types with `hasTypeRefInit == true`, the cctor is embedded *inside* the `_tri_b()` function body (TypeConverter.cs:1318-1326), so no cctor is lost by the split.
- Pass 1's missing null-check (line 533) is safe because `HasTypeRefInit == true` guarantees `GenerateLocalTypeRefInitCall` returns non-null.
- Regression test correctly exercises the exact crash scenario.

### One note (non-blocking):

**`ConvertForTests` has the same single-pass pattern** (RuntimeScopeManager.cs:355-368). If a framework test ever uses a static field with `ObservableCollection<T>`, it would hit the same ordering crash. Consider applying the same two-pass split there in a follow-up if framework tests start failing with similar symptoms. Not blocking since it's a test-only path and currently doesn't appear to hit this scenario.

### Inline note:
RuntimeScopeManager.cs:533 — No null-check here unlike Pass 2 (line 547). Safe because `HasTypeRefInit == true` guarantees non-null, but a defensive null-check would make the contract more obvious to future readers. Non-blocking.

PR #89, opened 2026-05-18
Fix: Razor skin templates write short name instead of absolute path in sourcesLong[] (#87)

### LINKED ISSUE #87: Fix: Razor skin templates write short name instead of absolute path in sourcesLong[]
## Problem

Razor `.skin.cshtml` templates were writing the short template name (e.g. `"AppShell"`) into `sourcesLong[]` of the generated source map instead of the absolute path to the `.skin.cshtml` file on disk.

This caused `SourceMap.Server` to return **404** for every Razor template source file because:
- `sourcesLong[i]` held `"AppShell"` (not a real path)
- The `AllowedSourceRoots` security check could never pass against a bare name
- DevTools could not load the original template source

## Root Cause

`TemplateIRBuilder.Build` used `templateName` (e.g. `"AppShell"`) as `Location.FileName` for all IR node locations. The full `.skin.cshtml` file path was available in `RazorTemplatingPlugin` but was dropped before reaching `TemplateIRBuilder`. Since `JSWriter` feeds `lastLocation.FileName` directly into `SourceMap.files[]` → `sourcesLong[]`, the short name propagated into the map.

## Fix

Thread the full source file path through the pipeline:

- `TemplateIRBuilder.Build` — add `string sourceFile = null`; compute `locationFileName = sourceFile ?? templateName`; use it for all `new Location(...)` constructions
- `RazorSkinCompiler.CompileToIR` — add `string sourceFile = null`; pass through to `Build`
- `RazorTemplatingPlugin` — pass `fileName` (from `GetResourceFileName`) as `sourceFile`

`templateName` retains its role as IR identity (`SkinTemplateNode.TemplateName`, scope identifiers). Only `Location.FileName` — which feeds source maps — changes to the absolute path.

## Regression Tests Added

7 new tests in `TemplateSourceMapTests` cover the `sourceFile` override through every node type:
- Root `SkinTemplateNode`
- `HtmlNode`
- `ExpressionBindingNode`
- `EventNode`
- `ConditionalNode` (`@if`)
- `LoopNode` (`@foreach`)
- Null fallback (omitting `sourceFile` still uses `templateName`)

gautam-achieveai:
Closes #87

## Summary

- Razor `.skin.cshtml` templates were writing the short template name (e.g. `"AppShell"`) into `sourcesLong[]` of the generated source map instead of the absolute disk path, causing `SourceMap.Server` to return 404 for every Razor template source file
- Root cause: `TemplateIRBuilder.Build` used `templateName` as `Location.FileName` for all IR nodes; the full `.skin.cshtml` path was available in `RazorTemplatingPlugin` but dropped before reaching the builder
- Fix: thread `sourceFile` (full path) through `CompileToIR` → `Build`; compute `locationFileName = sourceFile ?? templateName`; use for all `Location` constructions while keeping `templateName` as IR identity

## Changes

| File | Change |
|------|--------|
| `TemplateIRBuilder.cs` | Add `sourceFile` param; compute `locationFileName`; use for all `new Location(...)` |
| `RazorSkinCompiler.cs` | Add `sourceFile` param; pass through to `Build` |
| `RazorTemplatingPlugin.cs` | Pass `fileName` from `GetResourceFileName` as `sourceFile` |
| `TemplateSourceMapTests.cs` | 7 new regression tests covering `sourceFile` override through every IR node type |

## Test plan

- [x] All 830 tests pass (780 compiler + 50 E2E browser) — pre-commit hook verified
- [x] `TemplateSourceMapTests` — 7 new tests: root, HtmlNode, ExpressionBindingNode, EventNode, ConditionalNode, LoopNode, null fallback
- [x] Manual verification: `http://localhost:5005/sourcemap/TodoApp/Test/Framework/TodoApp/RazorTemplates/AppShell.skin.cshtml` returns 200 (was 404 before fix)
- [x] `sourcesLong[61]` in `TodoApp.map` is now `B:\sources\NScript\Test\Framework\TodoApp\RazorTemplates\AppShell.skin.cshtml` (was `"AppShell"`)

### DIFF
--- a/Sources/Compiler/RazorSkinParser/RazorSkinCompiler.cs
+++ b/Sources/Compiler/RazorSkinParser/RazorSkinCompiler.cs
@@ -23,7 +23,8 @@ namespace NScript.RazorSkin
         public static SkinTemplateNode CompileToIR(
             string templateName,
             string templateSource,
-            string[] additionalCSharpSources = null)
+            string[] additionalCSharpSources = null,
+            string sourceFile = null)
         {
             var log = Logger;
             var totalSw = Stopwatch.StartNew();
@@ -44,7 +45,7 @@ namespace NScript.RazorSkin
             // Phase 3: Build IR
             phaseSw.Restart();
             log.Debug("Phase {Phase} started for template {TemplateName}", "BuildIR", templateName);
-            var ir = TemplateIRBuilder.Build(templateName, preprocessed, parsed);
+            var ir = TemplateIRBuilder.Build(templateName, preprocessed, parsed, sourceFile);
             log.Debug("Phase {Phase} completed in {ElapsedMs}ms for template {TemplateName}", "BuildIR", phaseSw.ElapsedMilliseconds, templateName);
 
             // Phase 4: Roslyn analysis (refine classifications)
--- a/Sources/Compiler/RazorSkinParser/RazorTemplatingPlugin.cs
+++ b/Sources/Compiler/RazorSkinParser/RazorTemplatingPlugin.cs
@@ -191,7 +191,7 @@ namespace Sunlight.Framework.Observables
 
                             var ir = RazorSkinCompiler.CompileToIR(
                                 templateName, templateSource,
-                                additionalSources);
+                                additionalSources, fileName);
                             // Store under both short name and full resource name
                             // so [Skin("full.resource.name.skin.cshtml")] matches
                             _compiledIRs[templateName] = ir;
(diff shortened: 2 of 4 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** ✅ APPROVE

Clean, well-targeted fix for #87. The change correctly threads the full `.skin.cshtml` path through to `Location.FileName` so that `sourcesLong[]` entries in the source map resolve to actual disk paths instead of short template names.

**What's good:**
- Correct separation: `templateName` retains its identity role (SkinTemplateNode.TemplateName, scope identifiers); only `Location.FileName` changes
- Safe null fallback (`sourceFile ?? templateName`) ensures backward compat for callers without a path
- No breaking API changes -- optional params with `null` default
- Thorough test coverage: 7 regression tests covering all IR node types + the null fallback

**Minor observation (non-blocking):**
`WalkMethodBody`'s internal parameter is still named `templateName` (line 221) but now receives a file path value. A rename to `locationFileName` would improve clarity, but it's purely cosmetic and internal -- not worth a follow-up.

All 15 TemplateSourceMapTests pass.

PR #96, opened 2026-05-22
[Gautam's Dev bot] Fix WI-94: ErrorType arm in SymbolSerializer (sig-level)

### LINKED ISSUE #94: Compiler crash: declare-then-assign-in-try local crashes SymbolSerializer

gautam-achieveai:
[Gautam's Dev bot]

Closes #94

## Root cause

WI-93's `HasErrors` gate in `SerializationHelper.InjectIntoCompilation` is **body-only** — it skips method bodies that carry binding errors. A method whose body binds cleanly while its **signature** (param or return type) references an unresolved type slips past this gate.

The runtime path is:

1. `OnBoundExpressionGenerated` fires for the clean body.
2. `BoundAstToAstBase.GetMethodBody` calls `GetMethodSpecId(methodSymbol)`, which serializes the method signature **before** walking the body.
3. The unresolved param/return type is a `MissingMetadataTypeSymbol+TopLevel` (`Kind == ErrorType`).
4. `SymbolSerializer.Serialize(TypeSymbol)` falls through to the `throw new NotSupportedException(...)` at the non-`NamedType` arm, crashing the build instead of letting Roslyn surface its CS0246 diagnostic.

This was the actual trigger behind the "declare-then-assign-in-try local" symptom reported in the issue title — the title scenario itself is body-level and already covered by WI-93, but the sibling/signature-level case escapes.

## Fix

Add a dedicated `ErrorType` arm in `SymbolSerializer.Serialize(TypeSymbol)` that returns a placeholder `TypeSpecSer` (`Name = "<error>"`, `Module = null`). Roslyn then completes `Emit`, returns `Success == false`, and the underlying diagnostics surface normally. The placeholder is never consumed downstream because Stage 2 (Cs2Jsc) only runs when Stage 1 emit succeeds.

The fix is scoped to ~16 lines in `SymbolSerializer.cs`. The WI-93 body-level gate is unchanged (defense-in-depth). The existing throw for genuinely unsupported shapes (e.g. `FunctionPointerType`, future C# surface) is kept as a backstop with its diagnostic detail.

## Tests

`Test/Compiler/NScript.Csc.Lib.Test/ErrorTypeLeakTests.cs` exercises the signature-level path that WI-93 missed:

- `SignatureLevel_UnknownParameterType_DoesNotThrow` — bare param-type repro from the probe (`B2.DoSomething(NotDefinedType x)`).
- `SignatureLevel_UnknownReturnType_DoesNotThrow` — mirror for the return-type path.
- `SignatureLevel_GenericArgumentIsErrorType_DoesNotThrow` — `List<NotDefinedType>` (NamedType containing ErrorType recurses into the new arm).
- `SignatureLevel_ArrayOfErrorType_DoesNotThrow` — `NotDefinedType[]` (ArrayType element recurses into the new arm).
- `SignatureLevel_MixedWithHealthySibling_CapturesHealthyOnly` — healthy sibling method body still captured (per-callback gate).
- `TitleScenario_DeclareThenAssignInTry_DoesNotThrow` — locks the issue's named scenario in place.

All assert `ThrownException == null && Success == false && Diagnostics contains CS0246/CS0234/CS0117`.

Existing `CrossAssemblyInternalCrashTests` (WI-93) remain green — healthy-paths and per-callback semantics are unaffected.

## Test results

```
NScript.Csc.Lib.Test: Passed: 34, Skipped: 2, Failed: 0
  (CrossAssemblyInternalCrashTests + ErrorTypeLeakTests + neighbours)
```

Release build of `Sources/Compiler/NScript.Csc.Lib/` clean (0 errors, pre-existing CS8632 warnings only).

### DIFF
--- a/Sources/Compiler/NScript.Csc.Lib/SymbolSerializer.cs
+++ b/Sources/Compiler/NScript.Csc.Lib/SymbolSerializer.cs
@@ -211,6 +211,24 @@ namespace NScript.Csc.Lib
                 { PointedAtType = GetTypeSpecSer(((PointerTypeSymbol)type).PointedAtType) };
             }
 
+            if (type.Kind == SymbolKind.ErrorType)
+            {
+                // WI-94: An ErrorTypeSymbol reaches here when a method signature
+                // (param/return type) cannot be resolved (e.g. CS0246 unknown
+                // type). The body-level HasErrors gate in
+                // SerializationHelper.InjectIntoCompilation does not catch this
+                // because the owning method's body may bind cleanly. Returning
+                // a placeholder lets Roslyn finish emit; Emit() then reports
+                // Success=false with the underlying diagnostics, instead of
+                // crashing with NotSupportedException deep inside our hook.
+                return new TypeSpecSer
+                {
+                    Name = "<error>",
+                    Namespace = null,
+                    Module = null
+                };
+            }
+
             if (type.ContainingModule == null)
             {
                 return new TypeSpecSer
@@ -236,14 +254,14 @@ namespace NScript.Csc.Lib
 
             if (type.Kind != SymbolKind.NamedType)
             {
-                // WI-93: Surface enough context that the next regression in this
-                // arm is identifiable without an instrumentation cycle. Common
-                // ways to reach here include ErrorTypeSymbol (binding error),
-                // FunctionPointerType, and future C# surface (e.g. ref-struct
-                // shapes Roslyn introduces later). The primary HasErrors gate in
-                // SerializationHelper.InjectIntoCompilation suppresses the
-                // ErrorType case; this throw remains a backstop for genuinely
-                // unsupported shapes.
+                // WI-93/WI-94: Surface enough context that the next regression
+                // in this arm is identifiable without an instrumentation cycle.
+                // ErrorTypeSymbol is handled by the dedicated arm above; the
+                // body-level HasErrors path is filtered in
+                // SerializationHelper.InjectIntoCompilation. This throw remains
+                // a backstop for genuinely unsupported shapes such as
+                // FunctionPointerType or future C# surface (e.g. new ref-struct
+                // shapes Roslyn may introduce).
                 throw new NotSupportedException(
                     string.Format(
                         "NScript.Csc.Lib.SymbolSerializer.Serialize: unsupported TypeSymbol Kind='{0}', RuntimeType='{1}', FQN='{2}', ContainingType='{3}'.",
--- a/Test/Compiler/NScript.Csc.Lib.Test/ErrorTypeLeakTests.cs
+++ b/Test/Compiler/NScript.Csc.Lib.Test/ErrorTypeLeakTests.cs
@@ -1,0 +1,238 @@
+namespace NScript.Csc.Lib.Test
+{
+    using System;
+    using System.IO;
+    using System.Linq;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    /// <summary>
+    /// Regression coverage for WI-94: an <c>ErrorTypeSymbol</c> reaching
+    /// <c>SymbolSerializer.Serialize(TypeSymbol)</c> via the METHOD SIGNATURE
+    /// path (param/return type), not the method body. WI-93's
+    /// <c>HasErrors</c> gate in <c>SerializationHelper.InjectIntoCompilation</c>
+    /// is body-only, so it does not fire when a sibling method's body binds
+    /// cleanly but its signature references an undefined type. The fix is a
+    /// dedicated <c>ErrorType</c> arm in <c>Serialize(TypeSymbol)</c> that
+    /// returns a placeholder <c>TypeSpecSer</c> instead of throwing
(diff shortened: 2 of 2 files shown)


gautam-achieveai:
**[Gautam's review bot]** :white_check_mark: **APPROVE**

**Context**: Bug #94 — compiler crash when `SymbolSerializer.Serialize(TypeSymbol)` encounters an `ErrorTypeSymbol` via method *signature* (not body). WI-93's `HasErrors` gate is body-only, so it doesn't protect against unresolved types in parameter/return signatures when the method body itself binds cleanly.

**Fix correctness**: The dedicated `ErrorType` arm is correctly placed before the `ContainingModule == null` check and the `Kind != NamedType` backstop throw. The placeholder `TypeSpecSer(Name=\"<error>\", Module=null)` follows the existing `dynamic` pattern and is safe because emit always returns `Success=false` when an ErrorType is present, preventing Stage 2 from ever seeing it.

**Test coverage**: 6 new regression tests covering all relevant paths (param, return type, mixed-with-healthy-sibling, generic argument, array element, title scenario). All pass. Full test suite remains green (34/34, 2 skipped pre-existing).

**No issues found.** Clean, well-scoped fix with excellent comments explaining the rationale.

PR #86, opened 2026-05-18
Add secondary repo source root support to RepoLinkedSourceMaps

gautam-achieveai:
## Summary

Fixes #85

- Adds `SecondaryRepoRoot` / `SecondarySourceRoot` properties to `SourceMap` so files from a dependency repo are emitted as absolute `https://` URLs in `sources[]` rather than falling back to `AbsolutizeForSources()` (which produces `B$/sources/...` paths that combine nonsensically with the app's GitHub `sourceRoot`)
- Exploits the V3 source map spec: absolute URLs in `sources[]` are used verbatim by DevTools — `sourceRoot` is **not** prepended — so one `.map` can reference two distinct Git repos
- New compiler flags: `-secondarySourceRoot <url>` and `-secondaryRepoRoot <path>` wired through `ParseOptions` → `Builder` → `JSWriter` → `SourceMap`
- New MSBuild target `ComputeNScriptSecondaryRepoMetadata` auto-detects GitHub / ADO URL + commit SHA from the secondary repo using the same git exec + regex patterns as the primary `ComputeNScriptRepoMetadata`; exposes `NScriptSecondaryRepoRoot` / `NScriptSecondarySourceMapRoot` / `NScriptSecondaryCommitSha` / `NScriptSecondarySourceRepoOriginUrl` properties
- Feature generalised to any dependency repo (NScript framework, internal shared libraries, etc.) — not NScript-specific

## How to use

```xml
<!-- In your app's .csproj or Directory.Build.props -->
<PropertyGroup>
  <RepoLinkedSourceMaps>true</RepoLinkedSourceMaps>
  <!-- Path to the secondary repo's local checkout — SHA + remote URL auto-detected -->
  <NScriptSecondaryRepoRoot>$(MSBuildThisFileDirectory)..\NScript\</NScriptSecondaryRepoRoot>
</PropertyGroup>
```

For non-GitHub/ADO remotes, set `NScriptSecondarySourceMapRoot` manually (skips auto-detection).

## Test plan

- [x] `ToString_SecondaryRepoFile_EmitsAbsoluteUrl` — secondary repo file → absolute `https://` URL
- [x] `ToString_MixedBundle_PrimaryRelativeSecondaryAbsolute` — primary relative, secondary absolute in same map
- [x] `ToString_SecondaryRepoRoot_WithoutSecondarySourceRoot_FallsBackToLegacy` — graceful fallback when only one half configured
[...]

PR #98, opened 2026-05-22
Embed repo metadata in framework NuGets for zero-config source maps (#97)

### LINKED ISSUE #97: Embed repo metadata in NScript NuGets for zero-config multi-repo source maps
## Summary

NScript framework NuGet packages (`Sunlight.Framework`, `Sunlight.Framework.UI`,
`Sunlight.Framework.Data`, etc.) should carry their originating Git remote URL +
commit SHA inside the package so that a consuming app can produce repo-linked
source maps for framework code with **zero additional configuration**.

Today's state (after #85 / #86): `RepoLinkedSourceMaps=true` correctly links the
*app's own* sources to its GitHub origin via auto-detected `git` metadata, and a
secondary repo can be wired in via `NScriptSecondaryRepoRoot` / `…SourceMapRoot`
— but the consumer has to (a) clone the NScript repo somewhere on disk and
(b) point `NScriptSecondaryRepoRoot` at it. For most consumers that's
impractical: they install NScript via NuGet and have no checkout.

## Value

- **Developers consuming NScript via NuGet** can step into framework code in
  Chrome / Edge DevTools without cloning `achieveai/NScript` locally.
- **Production / CI builds** get correct framework source URLs automatically
  — no per-CI-job wiring of `NScriptSecondary*` properties.
- **Multiple NScript versions in flight** resolve to the right SHA per package,
  not whatever happens to be checked out on the dev box.

## Proposed Design (open for refinement)

1. **At pack time** in this repo, the SDK captures `git remote get-url origin`
   + `git rev-parse HEAD` of the NScript worktree and embeds them into each
   framework NuGet via **both** carriers:
   - A `buildTransitive/*.props` file that sets
     `NScriptPackageOriginUrl_<PackageId>` / `NScriptPackageCommitSha_<PackageId>`
     so consuming MSBuild can pick them up automatically.
   - An embedded resource (e.g. `$$NScriptPackageRepo$$`) inside the framework
     DLL itself, mirroring the existing `$$BstInfo$$` / `$$ResInfo$$` pattern,
     for tooling that reads the assembly without MSBuild context.
2. **At consumer build time**, when `RepoLinkedSourceMaps=true` and the SDK
   detects a referenced NScript framework assembly that ships repo metadata,
   it auto-resolves the corresponding `NScriptSecondarySourceMapRoot` (and
   appropriate `NScriptSecondaryRepoRoot` proxy) without any user property set.
3. The existing manual overrides (`NScriptSecondaryRepoRoot`, etc.) continue
   to work and take precedence — for the case where the consumer has a local
   checkout and wants to point at uncommitted changes.

## Acceptance Criteria

- [ ] NScript repo's pack pipeline captures origin URL + commit SHA and embeds
      them in every framework NuGet, via both `.props` and an embedded DLL
      resource.
- [ ] A consumer project with `<RepoLinkedSourceMaps>true</RepoLinkedSourceMaps>`
      and NO `NScriptSecondary*` properties produces a `.map` whose framework
      `sources[]` resolve to `https://raw.githubusercontent.com/achieveai/NScript/{sha}/Sources/Framework/...`.
- [ ] Manual `NScriptSecondaryRepoRoot` override still wins when set.
- [ ] Works for GitHub today; Azure DevOps follows the same shape (parity with
      existing `RepoUrlBuilder` coverage).
- [ ] Graceful fallback: if a package shipped without metadata (e.g. built
      outside a git checkout), consumer build emits a warning and falls back to
      legacy absolute-path behavior — never fails the build.
- [ ] Tests: pack a fixture package with embedded metadata, restore it from a
      test consumer, build, assert the `.map` `sources[]` URLs.
- [ ] Docs: update `docs/source-maps/repo-linked-source-maps.md` with the
      zero-config consumer scenario.

## Related

- #85 (closed) — original multi-repo source-root request
- #86 — added `NScriptSecondaryRepoRoot` plumbing (the foundation this builds on)
- #87 — fix for `sourcesLong[]` short-name leak

gautam-achieveai:
## Summary

Framework NuGet packages (`Mcqdb.NScript.Sunlight.Framework`, `…UI`, `…Data`, `…MsCorlib`, etc.) now ship their originating Git remote URL + commit SHA + build-machine worktree root inside the package, so a consumer can produce repo-linked source maps for framework code with only `<RepoLinkedSourceMaps>true</RepoLinkedSourceMaps>` set — no `NScriptSecondary*` properties required, no need to clone the NScript repo locally.

Built on top of PR #86's `NScriptSecondaryRepoRoot` machinery — does not rewrite it. The existing regex URL builder is reused unchanged; this PR just auto-feeds it the values it needs.

### Pack time (NScript repo)

`Sources/Framework/NScript.PackageMetadata.targets` (new, imported from `Sources/Framework/Directory.Build.props`) captures git metadata once per framework project and emits TWO carriers, per the issue's acceptance criteria:

1. **`buildTransitive/<PackageId>.props`** packed into the NuGet — MSBuild auto-imports this on the consumer. Sets per-id properties `NScriptPackageOriginUrl_<SafeId>` / `…CommitSha_<SafeId>` / `…RepoRoot_<SafeId>`, and appends to a shared `@(NScriptPackagesWithRepoMetadata)` item group for enumeration.
2. **`$$NScriptPackageRepo$$` resource** embedded in the framework DLL, mirroring the existing `$$BstInfo$$` / `$$ResInfo$$` pattern. Carries the same three values in a simple key=value text format.

### Consumer side

`Sources/Compiler/NScript.Sdk/Sdk/Sdk.targets` gains `ComputeNScriptPackageMetadataFromReferences`, a tiny target that runs before the existing `ComputeNScriptSecondaryRepoMetadata`. It picks the first entry from `@(NScriptPackagesWithRepoMetadata)` and feeds `NScriptSecondaryRepoRoot` / `…CommitSha` / `…SourceRepoOriginUrl`. The existing regex URL builder then computes `NScriptSecondarySourceMapRoot` with zero changes to its logic.

Skipped entirely when `NScriptSecondaryRepoRoot` is set manually — **manual override always wins** (acceptance criterion #3).

### Why this works without a local checkout

`SourceMap.GetSourceEntry` matches source-file paths against `NScriptSecondaryRepoRoot` via `TryRebaseToRepoRoot` — a **string-prefix match**. The "repo root" doesn't need to exist on the consumer's disk; it just needs to match the prefix baked into the framework's `\$\$BstInfo\$\$` resource at framework-build time, which is exactly what we now embed.

## Key Decisions

- **Two carriers** — `<EmbeddedResource>` plus `<None Pack>`. csc honors the resource via the standard `-resource:` switch (no NScript csc fork changes needed).
- **First-wins on consumer** — all 11 framework packages built together share identical metadata, so picking the first entry is safe. The auditing `Message` logs every detected package.
- **Properties inside the target body** — initial draft put file-path properties in a top-level `PropertyGroup`; that evaluated before `PackageId` and `IntermediateOutputPath` were set, collapsing paths to `.repo.bin` in the project root. Fixed by deferring evaluation until target run time.
- **`BeforeTargets="_GenerateCompileInputs;CoreCompile;GenerateNuspec"`** — the SDK's `_GenerateCompileInputs` snapshots `@(EmbeddedResource)` into `@(_CoreCompileResourceInputs)`, which is what csc actually reads. Without the explicit ordering, dynamic resource additions never reach the compiler.
- **`Type=Non-Resx` + `WithCulture=false`** on the `<EmbeddedResource>` — required to satisfy `_GenerateCompileInputs`'s filtering condition.
- **BOM-tolerant parser** — `WriteLinesToFile` emits a UTF-8 BOM that broke the first-key match; reader strips a leading `U+FEFF`.

## Files Changed

| File | Action | Description |
|------|--------|-------------|
| `Sources/Framework/NScript.PackageMetadata.targets` | Add | Pack-time target that captures git metadata, writes `.bin` + `.buildTransitive.props`, adds `<EmbeddedResource>` and `<None Pack>` items |
| `Sources/Framework/Directory.Build.props` | Modify | Import the new targets file so all 11 framework projects pick it up |
| `Sources/Compiler/NScript.Sdk/Sdk/Sdk.targets` | Modify | Add `ComputeNScriptPackageMetadataFromReferences` target |
| `Sources/Compiler/NScript.CLR/PackageRepoMetadata.cs` | Add | Mono.Cecil-based reader for the `\$\$NScriptPackageRepo\$\$` resource |
| `Test/Compiler/NScript.CLR.Test/PackageRepoMetadataTests.cs` | Add | 23 parser/reader tests (BOM, CRLF, duplicate keys, malformed input, real-DLL E2E) |
| `Test/Compiler/NScript.CLR.Test/PackageMetadataTargetsTests.cs` | Add | 8 structural-invariant tests on the `.targets` files |
| `docs/source-maps/repo-linked-source-maps.md` | Modify | New "Zero-Config NuGet Consumer Scenario" section + properties-reference rows for the new properties |

## Test Plan

- [x] `dotnet build NScript_Full.sln -c Release` — 0 errors
- [x] `dotnet test Test/Compiler/NScript.CLR.Test/...` — 120/120 pass (97 pre-existing + 23 new)
- [x] `dotnet test Test/Compiler/SourceMap.Test/...` — 88/88 pass (no regression to PR #86)
- [x] `dotnet test Test/Compiler/NScript.Utils.Test/...` — 33/33 pass
- [x] Pre-commit hook full test matrix — 860 passed, 0 failed (810 compiler + 50 e2e)
- [x] Manual nupkg inspection — `buildTransitive/Mcqdb.NScript.Sunlight.Framework.props` present with per-id properties + `@(NScriptPackagesWithRepoMetadata)` entry; DLL contains `\$\$NScriptPackageRepo\$\$` resource with the three-key payload.
- [x] End-to-end test in `PackageRepoMetadataTests.TryReadFromAssembly_RealFrameworkAssembly_RoundTripsMetadata` reads the real framework DLL via Cecil and confirms a populated metadata struct with a 40-char SHA.
- [ ] Smoke test in a real downstream consumer (manual, post-merge) — set only `<RepoLinkedSourceMaps>true</RepoLinkedSourceMaps>` in a fresh consumer project, build, inspect the produced `.map`'s `sources[]` for `https://raw.githubusercontent.com/achieveai/NScript/<sha>/...` entries.

Closes #97.

### DIFF
--- a/Sources/Compiler/NScript.CLR/PackageRepoMetadata.cs
+++ b/Sources/Compiler/NScript.CLR/PackageRepoMetadata.cs
@@ -1,0 +1,227 @@
+//-----------------------------------------------------------------------
+// <copyright file="PackageRepoMetadata.cs" company="">
+//     Copyright (c) . All rights reserved.
+// </copyright>
+//-----------------------------------------------------------------------
+
+namespace NScript.CLR
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Text;
+    using Mono.Cecil;
+
+    /// <summary>
+    /// Metadata describing the Git repository that produced an NScript framework NuGet
+    /// package, captured at pack time and embedded into the package's primary assembly as
+    /// a <c>$$NScriptPackageRepo$$</c> resource (mirroring the existing
+    /// <c>$$BstInfo$$</c> / <c>$$ResInfo$$</c> pattern). Consumed by the consumer-side
+    /// SDK target to auto-resolve secondary source-map URLs without requiring the
+    /// consumer to clone the NScript repo locally (work item #97).
+    /// </summary>
+    /// <remarks>
+    /// The primary delivery path for this data is the package's
+    /// <c>buildTransitive/&lt;PackageId&gt;.props</c> file, which MSBuild auto-imports
+    /// when the package is referenced. This reader exists for MSBuild-less tooling
+    /// (e.g. a hypothetical standalone <c>cs2jsc</c> invocation outside the SDK) that
+    /// needs to recover the same values directly from the assembly via Mono.Cecil.
+    ///
+    /// Wire format: UTF-8 text, one <c>key=value</c> pair per line, LF or CRLF
+    /// line endings tolerated. The canonical keys are <c>originUrl</c>,
+    /// <c>commitSha</c>, and <c>repoRoot</c>. Unknown keys are silently ignored so
+    /// future schema additions remain backward-compatible. Blank lines and lines
+    /// without an <c>=</c> separator are skipped.
+    /// </remarks>
+    public sealed class PackageRepoMetadata
+    {
+        /// <summary>
+        /// Name of the embedded resource that carries the serialized metadata. The
+        /// `$$` framing mirrors <c>$$BstInfo$$</c> / <c>$$ResInfo$$</c> so the pattern
+        /// is consistent across NScript-emitted resources.
+        /// </summary>
+        public const string ResourceName = "$$NScriptPackageRepo$$";
+
+        private PackageRepoMetadata(string originUrl, string commitSha, string repoRoot)
+        {
+            this.OriginUrl = originUrl;
+            this.CommitSha = commitSha;
+            this.RepoRoot = repoRoot;
+        }
+
+        /// <summary>
+        /// Git remote URL captured at pack time (output of
+        /// <c>git remote get-url origin</c>). May contain credentials in the
+        /// <c>https://user:token@host/...</c> form on machines where developers have
+        /// configured them locally; callers that surface this value into logs MUST
+        /// apply the <c>://creds@</c> redaction used elsewhere in the SDK.
+        /// </summary>
+        public string OriginUrl { get; }
+
+        /// <summary>
+        /// Commit SHA captured at pack time (output of <c>git rev-parse HEAD</c>).
+        /// The exact SHA of the framework sources baked into this assembly.
+        /// </summary>
+        public string CommitSha { get; }
+
+        /// <summary>
+        /// Worktree root captured at pack time (output of
+        /// <c>git rev-parse --show-toplevel</c>). This is a BUILD-MACHINE absolute
+        /// path; it does not exist on the consumer's machine. Its purpose is to
+        /// match the path prefix baked into the assembly's <c>$$BstInfo$$</c>
+        /// source-file references so that
+        /// <c>OwaSourceMapper.SourceMap.TryRebaseToRepoRoot</c> can rebase framework
+        /// sources to repo-relative form.
+        /// </summary>
+        public string RepoRoot { get; }
+
+        /// <summary>
+        /// Reads the <c>$$NScriptPackageRepo$$</c> resource from <paramref name="assemblyPath"/>,
+        /// if present, and parses it into a <see cref="PackageRepoMetadata"/> instance.
+        /// </summary>
+        /// <param name="assemblyPath">Path to a framework NuGet's primary assembly.</param>
+        /// <returns>
+        /// Parsed metadata, or <c>null</c> when the resource is absent, malformed, or
+        /// the file cannot be opened. Callers should treat <c>null</c> as "no metadata
+        /// available" — the consumer build proceeds with legacy local-path source maps.
+        /// </returns>
+        public static PackageRepoMetadata TryReadFromAssembly(string assemblyPath)
+        {
+            if (string.IsNullOrEmpty(assemblyPath) || !File.Exists(assemblyPath))
+            {
+                return null;
+            }
+
+            ModuleDefinition module;
+            try
+            {
+                // No symbol reading, no metadata resolver gymnastics — we just want the
+                // resource bytes. Mono.Cecil keeps the file open for the lifetime of the
+                // ModuleDefinition, so we read into memory and dispose immediately.
+                module = ModuleDefinition.ReadModule(assemblyPath);
+            }
+            catch (BadImageFormatException) { return null; }
+            catch (IOException) { return null; }
+            catch (UnauthorizedAccessException) { return null; }
+
+            using (module)
+            {
+                var resource = module.Resources
+                    .OfType<EmbeddedResource>()
+                    .FirstOrDefault(r => string.Equals(r.Name, ResourceName, StringComparison.Ordinal));
+
+                if (resource == null)
+                {
+                    return null;
+                }
+
+                byte[] bytes;
+                try
+                {
+                    bytes = resource.GetResourceData();
+                }
+                catch (IOException) { return null; }
+
+                return TryParse(bytes);
+            }
+        }
+
+        /// <summary>
+        /// Parses the raw resource bytes (UTF-8 key=value lines) into a metadata instance.
+        /// Public so MSBuild-less tooling that fetches the resource bytes via another
+        /// path (e.g. a CI script that opens a <c>.nupkg</c> directly) can reuse the
+        /// canonical parser instead of re-implementing the format.
+        /// </summary>
+        /// <param name="bytes">Raw resource payload.</param>
+        /// <returns>
+        /// Parsed metadata, or <c>null</c> when any of the three required keys
+        /// (<c>originUrl</c>, <c>commitSha</c>, <c>repoRoot</c>) is missing or empty.
+        /// </returns>
+        public static PackageRepoMetadata TryParse(byte[] bytes)
+        {
+            if (bytes == null || bytes.Length == 0)
+            {
+                return null;
+            }
+
+            string text;
+            try
+            {
+                text = Encoding.UTF8.GetString(bytes);
+            }
+            catch (ArgumentException)
+            {
+                return null;
+            }
+
+            return TryParse(text);
+        }
+
+        /// <summary>
+        /// Text-level overload of <see cref="TryParse(byte[])"/>. Same contract — returns
+        /// <c>null</c> for any input that lacks a non-empty <c>originUrl</c>,
+        /// <c>commitSha</c>, and <c>repoRoot</c> pair.
+        /// </summary>
+        public static PackageRepoMetadata TryParse(string text)
+        {
+            if (string.IsNullOrEmpty(text))
+            {
+                return null;
+            }
+
+            // MSBuild's WriteLinesToFile task writes UTF-8 WITH a BOM (U+FEFF
+            // prefix). Without stripping it, the first key would parse as
+            // "﻿originUrl" instead of "originUrl" and the required-key
+            // check below would return null. Strip a single leading BOM.
+            if (text.Length > 0 && text[0] == '﻿')
+            {
+                text = text.Substring(1);
+            }
+
+            var values = new Dictionary<string, string>(StringComparer.Ordinal);
+
+            // Splitting on '\n' and trimming '\r' tolerates both LF and CRLF without
+            // pulling in StringReader. Blank lines and `key`-without-`=` lines are
+            // skipped silently — the format is intentionally forgiving so future
+            // schema additions remain backward-compatible.
+            foreach (var rawLine in text.Split('\n'))
+            {
+                var line = rawLine.TrimEnd('\r').Trim();
+                if (line.Length == 0)
+                {
+                    continue;
+                }
+
+                int eq = line.IndexOf('=');
+                if (eq <= 0)
+                {
+                    continue;
+                }
+
+                var key = line.Substring(0, eq).Trim();
+                var value = line.Substring(eq + 1).Trim();
+                if (key.Length == 0 || value.Length == 0)
+                {
+                    continue;
+                }
+
+                // Last-wins on duplicate keys — matches MSBuild PropertyGroup semantics.
+                values[key] = value;
+            }
+
+            values.TryGetValue("originUrl", out var originUrl);
+            values.TryGetValue("commitSha", out var commitSha);
+            values.TryGetValue("repoRoot", out var repoRoot);
+
+            if (string.IsNullOrEmpty(originUrl)
+                || string.IsNullOrEmpty(commitSha)
+                || string.IsNullOrEmpty(repoRoot))
+            {
+                return null;
+            }
+
+            return new PackageRepoMetadata(originUrl, commitSha, repoRoot);
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of PR #98 — Embed repo metadata in framework NuGets for zero-config source maps (#97)

---

## Context

**Work Item #97**: Embed repo metadata in NScript NuGets for zero-config multi-repo source maps.

This PR adds a pack-time MSBuild target that captures git metadata (origin URL, commit SHA, worktree root) and ships it in framework NuGet packages via two carriers: a `buildTransitive/<PackageId>.props` file (auto-imported by MSBuild) and a `$$NScriptPackageRepo$$` embedded resource (for MSBuild-less tooling). The consumer-side SDK auto-resolves secondary source-map URLs from the first `@(NScriptPackagesWithRepoMetadata)` entry, eliminating the need for manual `NScriptSecondary*` configuration.

---

## Findings

### MEDIUM — Unredacted origin URL embedded in NuGet artifacts

`NScript.PackageMetadata.targets` lines 130 and 160 write the raw `$(_NScriptPkgOrigin)` into both the `.repo.bin` resource and the `buildTransitive` .props file. The redacted form (`$(_NScriptPkgOriginRedacted)`) is only used for the `<Message>` log output.

If a developer (or CI machine) has credentials embedded in their git remote URL (e.g., `https://token@github.com/achieveai/NScript.git`), those credentials get baked into the published NuGet package. While the `PackageRepoMetadata.OriginUrl` doc comment correctly warns callers to redact before logging, the real risk is **the credential being persisted in a published artifact** — not just logged.

**Suggested mitigation**: Apply the same `://[^/@]+@` regex replacement to the value written into the two carrier files, not just the log message. The consumer-side regex builder only cares about the host/path portion of the URL anyway — credentials are irrelevant for URL construction.

### LOW — Version jump from 1.1.4 to 1.1.6

The root `NScriptPackageVersion` jumps from 1.1.4 to 1.1.6, and Cs2Jsc from 1.1.3 to 1.1.6. Was 1.1.5 consumed by something unpublished, or is this intentional alignment?

---

## Positive Observations

1. **Robust parser**: BOM handling, CRLF/LF tolerance, duplicate-key last-wins semantics, forward-compatible unknown-key skipping — well-thought-out wire format.
2. **Structural invariant tests** (`PackageMetadataTargetsTests.cs`): Clever approach — asserting MSBuild XML structure catches accidental edits without needing full pack/restore cycles.
3. **Graceful degradation**: Missing git metadata -> warning + skip (no build failure). Empty item group -> no-op. Manual override -> auto-resolve skipped. All fallback paths are clean.
4. **Follows existing patterns**: `$$NScriptPackageRepo$$` naming mirrors `$$BstInfo$$`/`$$ResInfo$$`, and the key=value format is simple and extensible.
5. **Good documentation**: The new "Zero-Config NuGet Consumer Scenario" section covers debugging/override/fallback scenarios.
6. **BeforeTargets + Type/WithCulture metadata**: The `_GenerateCompileInputs` ordering constraint is documented inline — this MSBuild edge case is a common pitfall.

---

## Verdict

**Approve with comment** — the credential-leakage concern is worth addressing before this ships to a public feed, but the overall implementation is solid, well-tested, and follows established patterns cleanly.

Sources/Framework/NScript.PackageMetadata.targets
```diff
@@ -0,0 +1,219 @@
...
+
+    <!--
+      .repo.bin payload — simple key=value text, one pair per line. Keep this
+      format stable: PackageRepoMetadata.cs (NScript.CLR) parses it, and a
+      schema change here means a schema change there. New keys can be appended;
+      unknown keys are tolerated by the parser.
+    -->
+    <ItemGroup Condition="'$(_NScriptPkgMetadataAvailable)' == 'true'">
+      <_NScriptPkgBinLines Include="originUrl=$(_NScriptPkgOrigin)" />
+      <_NScriptPkgBinLines Include="commitSha=$(_NScriptPkgSha)" />
```
gautam-achieveai, line 143: **[Gautam's review bot]** :yellow_circle: **Credential leakage risk**: This writes the raw, unredacted origin URL into the `.repo.bin` payload that gets embedded in the DLL and shipped in the NuGet package. If a developer has credentials in their git remote (e.g., `https://PAT@github.com/...`), those credentials end up in the published artifact.

The consumer-side regex builder only needs `github.com/{owner}/{repo}` — credentials are stripped by the regex pattern anyway. Consider applying the same `://[^/@]+@` replacement here (using `$(_NScriptPkgOriginRedacted)`) instead of `$(_NScriptPkgOrigin)`.

Same issue applies to line 160 (the buildTransitive .props carrier).

Sources/Compiler/NScript.CLR/PackageRepoMetadata.cs
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        /// <summary>
+        /// Git remote URL captured at pack time (output of
+        /// <c>git remote get-url origin</c>). May contain credentials in the
+        /// <c>https://user:token@host/...</c> form on machines where developers have
+        /// configured them locally; callers that surface this value into logs MUST
+        /// apply the <c>://creds@</c> redaction used elsewhere in the SDK.
+        /// </summary>
+        public string OriginUrl { get; }
```
gautam-achieveai, line 72: **[Gautam's review bot]** The doc comment correctly warns that callers must redact before logging — but the bigger risk is that this value is baked into a _published NuGet package_ (via `NScript.PackageMetadata.targets` line 131). The warning should also mention the artifact-persistence risk, not just the logging risk. Consider whether the pack-time target should strip credentials _before_ embedding.

Sources/Framework/NScript.PackageMetadata.targets
gautam-achieveai: [gautam-achieveai's bot] Fixed in bbecf743. Both carriers now embed `$(_NScriptPkgOriginRedacted)` instead of the raw `$(_NScriptPkgOrigin)`:

- `_NScriptPkgBinLines` (`originUrl=...` line in the `.repo.bin` payload that becomes the `$$NScriptPackageRepo$$` resource)
- `_NScriptPkgPropsLines` (the `NScriptPackageOriginUrl_<SafeId>` property and the `<OriginUrl>` child of `@(NScriptPackagesWithRepoMetadata)` in the `buildTransitive/<PackageId>.props` carrier)

The `://[^/@]+@` -> `://***@` regex was already being computed at line 115 of the targets file — only the consumers of the property had to flip from raw to redacted. For origin URLs without embedded credentials the regex is a no-op, so the change is a safe drop-in. Updated the surrounding comment to call out that the redaction is now load-bearing for the embedded artifact (not just for build-log output).

A new structural-invariant test `FrameworkTargets_EmbedsRedactedOriginUrlNotRaw` in `Test/Compiler/NScript.CLR.Test/PackageMetadataTargetsTests.cs` walks every `_NScriptPkgBinLines` / `_NScriptPkgPropsLines` `Include` value and rejects any that reference raw `$(_NScriptPkgOrigin)`, so the leak cannot regress silently.

Sources/Compiler/NScript.CLR/PackageRepoMetadata.cs
gautam-achieveai: [gautam-achieveai's bot] Fixed in bbecf743. The fix is applied at the source — the pack-time target in `Sources/Framework/NScript.PackageMetadata.targets` now writes `$(_NScriptPkgOriginRedacted)` into BOTH carriers (the `.repo.bin` resource payload AND the `buildTransitive/<PackageId>.props` file), so the value never reaches a published NuGet in its raw form.

The `OriginUrl` doc comment in `PackageRepoMetadata.cs` has been rewritten to reflect the new invariant:

- Previously: warned that callers logging the value MUST apply the `://creds@` redaction.
- Now: states positively that the value is pre-redacted at pack time (with a pointer to `_NScriptPkgOriginRedacted` and the `://[^/@]+@` regex) and is safe to log / ship to other tooling without further filtering. Added a `<remarks>` note that future producers emitting this value into a published artifact MUST preserve the pre-redaction invariant, with a pointer to the new structural test that guards the MSBuild side.

Parser coverage for the redacted form is now locked in by `TryParse_RedactedOriginUrl_PreservedAsIs` in `Test/Compiler/NScript.CLR.Test/PackageRepoMetadataTests.cs` (asserts `https://***@github.com/...` round-trips through the parser unchanged).

gautam-achieveai:
[gautam-achieveai's bot] Follow-up on the review summary (#98):

**MEDIUM — Credential leakage into published NuGet artifacts:** Fixed in bbecf743.

- The pack-time target in `Sources/Framework/NScript.PackageMetadata.targets` now embeds `$(_NScriptPkgOriginRedacted)` (the value already produced by the `://[^/@]+@` -> `://***@` regex) into BOTH carriers — `_NScriptPkgBinLines` (the `.repo.bin` payload that becomes the `$$NScriptPackageRepo$$` resource) and `_NScriptPkgPropsLines` (the `buildTransitive/<PackageId>.props` file). Raw origin URLs with embedded credentials can no longer ship inside a published package.
- The `OriginUrl` XML doc comment in `Sources/Compiler/NScript.CLR/PackageRepoMetadata.cs` has been rewritten from a warning ("callers must redact before logging") to a positive statement of the new invariant ("value is pre-redacted at pack time"), with a defensive `<remarks>` block reminding future producers to preserve it.
- Two new tests guard the invariant: `FrameworkTargets_EmbedsRedactedOriginUrlNotRaw` (structural — rejects any `_NScriptPkgBinLines` / `_NScriptPkgPropsLines` Include that references the raw `$(_NScriptPkgOrigin)`) and `TryParse_RedactedOriginUrl_PreservedAsIs` (parser — locks the `https://***@host/...` round-trip).
- Full CLR.Test suite: 28/28 passing. Full pre-commit matrix: 862/862 passing.

See thread #98 (discussion_r3292064933) and thread #98 (discussion_r3292065164) for the per-thread replies.

**LOW — Version jump 1.1.4 → 1.1.6 without 1.1.5:** This PR does NOT change `NScriptPackageVersion`. The bump happened on master before this branch was cut:

- `b0a4cc6e` "Bump NScriptPackageVersion to 1.1.6 (Cs2Jsc included)" landed on master ahead of this work.
- `git diff master -- Directory.Build.props` on the PR head returns empty output, confirming the property file is unchanged on this branch.

Per the babysit instructions for this iteration, no version-bump action is needed here. If 1.1.5 needs to be backfilled as an explicit milestone, that's a separate concern to track on master directly.

gautam-achieveai (commented):
**[Gautam's review bot]** Incremental re-review (iteration 3, commit `bbecf743`)

## Previous MEDIUM — Credential leakage: RESOLVED

The fix is thorough:
1. Both carriers now embed the redacted origin instead of raw
2. Doc comment updated to document the pre-redaction invariant
3. Structural regression test (`FrameworkTargets_EmbedsRedactedOriginUrlNotRaw`) guards both carriers
4. Parser compatibility test confirms `https://***@host/...` format is handled
5. Clear inline MSBuild comments explain why raw property is retained (regex input only - never persisted)

All 9 `PackageMetadata*` tests pass. No new issues found.

**Verdict: APPROVE**

PR #100, opened 2026-05-29
NScript #99: Sunlight log-ingestion pipeline (controller + WS + failover sink)

### LINKED ISSUE #99: SunlightTestAdapter: bridge browser test logs to structured JSONL file
## Summary
Bridge structured logs from in-browser Sunlight test runs to disk via a real HTTP/WebSocket ingestion pipeline, mapping each event through `Microsoft.Extensions.Logging` so any backend (Serilog file sink, console, App Insights, etc.) can consume them with full structured-properties fidelity.

The existing `HttpLogSink` (PR #24) already ships log batches to a configurable endpoint. This issue adds (a) the **server-side ingest endpoint** those batches POST to, and (b) a **WebSocket peer transport** with delivery validation, plus a **failover sink** that auto-prefers WS when connected.

Complements #10 (core pipeline JSONL logging) — same idea, different runtime layer.

## Architecture (locked via planning conversation)

**Server side — new shared project `Sources/Framework/Sunlight.Logging.Server/`**
- `ILogIngestionService.IngestAsync(LogEnvelope, CancellationToken)` — single ingestion method. Both transports converge here. Returns the set of event ids successfully written via `Microsoft.Extensions.Logging`.
- `LogController` (`[ApiController, Route("_log")]`) — `[HttpPost]` deserializes the envelope and calls `IngestAsync`.
- WebSocket endpoint at `/_log/ws` — uses `app.UseWebSockets()`, reads frames, deserializes envelope, calls `IngestAsync`, sends back `{"ackIds":[...]}` per-event ACK.
- Each `LogEvent`'s `props` + correlation fields (`traceId`, `spanId`, `parentSpanId`, `actionId`, `depth`) become `BeginScope` state so Serilog-or-whatever destructures them as structured properties.
- DI helpers: `services.AddSunlightLogIngestion()` and `app.MapSunlightLogIngestion(prefix)`.

**Adapter side — `SunlightTestAdapter`**
- Boots a minimal `WebApplication` on `localhost:0` during each test run; tears down on browser close.
- Configures `ILoggerFactory` with the same Serilog (or MEL) sinks the consumer wants (e.g. JSONL file).
- Injects the resolved endpoint URL into the bundle via a `window.__nscriptLogEndpoint` global on the synthetic index.html shim.
- New `Settings.LogEndpoint` (optional) — when set, enables the host; when unset, feature stays off.

**Client side — `Sources/Framework/Sunlight.Framework/Logging/`**
- `LogEvent` gains an `Id` field (per-event correlation ID). `LogJsonBuilder` emits `"id":"..."` in both transports — keeps wire shape symmetric.
- New `IWebSocketLogTransport` interface: `IsConnected`, `Send(payload)`, `AcksReceived` event.
- New `WindowWebSocketTransport` concrete impl using `[Script]` bridges to native `WebSocket`.
- New `WebSocketLogSink` — symmetric to `HttpLogSink`, tracks unacked events, retransmits on `ackTimeoutMs`.
- New `FailoverLogSink` wrapping both: routes to WS when `IsConnected`, falls back to HTTP otherwise; drains pending WS queue to HTTP on disconnect.

**Three client modes (per requirement)**
1. URL only → `new HttpLogSink(endpoint, ...)` (already shipped in #24)
2. WS only → `new WebSocketLogSink(transport, ...)`
3. Both with failover → `new FailoverLogSink(new WebSocketLogSink(...), new HttpLogSink(...))`

## Value
- Structured browser-side logs surface in the same Serilog JSONL stream `Microsoft.Extensions.Logging` writes for the server — single query target via DuckDB / Seq / App Insights.
- WS delivery validation (per-event ACK + retransmit) closes the data-loss gap the existing fire-and-forget HTTP path has.
- Failover sink means a flaky WS connection silently degrades to HTTP rather than dropping logs.
- `Microsoft.Extensions.Logging` decoupling lets consumers plug in whatever provider (Serilog, console, OTel) without coupling the framework to one backend.

## Acceptance Criteria
- [ ] New project `Sunlight.Logging.Server` builds and packages cleanly.
- [ ] `ILogIngestionService.IngestAsync` maps every `LogEvent.props` + correlation field to `BeginScope` state so they appear as structured properties in Serilog output.
- [ ] HTTP POST to `/_log` and WS frame to `/_log/ws` both produce identical `Microsoft.Extensions.Logging` output for the same envelope.
- [ ] WS handler replies with `{"ackIds":[...]}` listing the event ids successfully ingested for that envelope.
- [ ] `LogEvent` carries an `Id`; HTTP and WS envelopes both include it; server de-dupes by id.
- [ ] `WebSocketLogSink` retransmits any event whose id has not been acked within `ackTimeoutMs`.
- [ ] `FailoverLogSink` routes to WS when transport reports `IsConnected`, falls back to HTTP otherwise; drains pending WS queue to HTTP on mid-batch disconnect.
- [ ] `SunlightTestAdapter` boots the host when `Settings.LogEndpoint` is set; bundle receives the endpoint via `window.__nscriptLogEndpoint`.
- [ ] When `Settings.LogEndpoint` is unset, behavior is unchanged from today.
- [ ] Existing `page.Console` `[browser:log]` and `page.PageError` `[browser:pageerror]` forwarding to `IMessageLogger` is preserved unchanged.
- [ ] Unit tests cover: ingestion service, HTTP controller, WS handler, per-event ACK semantics, `WebSocketLogSink` retransmit path, `FailoverLogSink` failover.
- [ ] Integration test: run a test bundle that emits structured logs, assert they appear in the configured Serilog JSONL output with correct levels, categories, props, and correlation fields.

## Notes
- Bundle-side bootstrap must call `Logger.ClearSinks()` before any code that might log, to prevent the lazy `ConsoleSink` from racing in. Custom sinks must not call back into `Logger.*` (the re-entrancy guard would drop the nested call).
- `[Conditional("DEBUG")]` strips Trace/Debug in Release — test bundles must be Debug-compiled to capture those levels.
- Files of interest:
  - Server (new): `Sources/Framework/Sunlight.Logging.Server/`
  - Adapter: `Sources/Compiler/SunlightTestAdapter/Runner.cs`, `Executor.cs`, `Settings.cs`
  - Client framework: `Sources/Framework/Sunlight.Framework/Logging/` — extend `LogEvent`, `LogJsonBuilder`; add `IWebSocketLogTransport`, `WindowWebSocketTransport`, `WebSocketLogSink`, `FailoverLogSink`

**Priority:** Medium

---
*Note: this issue body was rewritten on 2026-05-25 after a planning conversation pivoted the architecture from marker-prefix-on-page.Console to a controller-based ingestion pipeline. See the next BOT-PLAN comment for the implementation plan.*

gautam-achieveai:
Fixes #99

## Summary

Bridges browser-side Sunlight structured logs into a server-side
`Microsoft.Extensions.Logging` pipeline. Implements the v2 plan
([#99 comment](https://github.com/achieveai/NScript/issues/99#issuecomment-4572197674))
plus the approved dogfood/E2E addendum
([ack comment](https://github.com/achieveai/NScript/issues/99#issuecomment-4572204398)).

## Key decisions (locked via HITL)

- **D1** In-process Kestrel + ASP.NET Core controller (not Playwright route interception).
- **D2-rev** New top-level `Sources/Runtime/` for server-runtime libraries (conceptually distinct from `Sources/Framework/` browser-runtime and `Sources/Compiler/` build-time).
- **D3-rev** Callback-based / BYOWS `WebSocketLogSink` — consumer owns the WS, sink takes `Func<bool> isConnected` + `Action<string> sendPayload` closures.
- **D7** Server `ILogIngestionService` is the public surface; `WebSocketLogProtocol.HandleFrameAsync` is a stateless BYOWS helper.
- **D8** Same-port unification: Kestrel serves both static assets and `/_log`+`/_log/ws`. Browser navigates to `http://127.0.0.1:N/index.html`.
- **D9** `maxRetry` per-event retransmit counter; `RetryExhausted` event lets `FailoverLogSink` drain to HTTP.
- **D10** `[ThreadStatic] _ingesting` guard prevents MEL re-entrancy.
- **D11** `LogEvent.Id` is per-emit runtime-unique (not source-stable). Required for WS per-event ACK targeting.
- **D12** Kestrel only starts when `Settings.LogEndpoint != null`. When unset, today's Playwright-route static-serving path is preserved verbatim — zero behavioral change for existing consumers.

## 18 ordered steps (each a separate commit)

1. `LogEvent.Id` field + `Logger.GenerateEventId()` + `LogJsonBuilder` emit, with 9 new LogEventIdTests covering all five levels, distinct same-call-site ids, NamedLogger path, single-event and envelope JSON shape.
2. `Sources/Runtime/` scaffold + `Sunlight.Logging.Server.csproj` (net8.0, AspNetCore.App, packaged as `Mcqdb.NScript.Sunlight.Logging.Server`).
3. POCOs: `LogEnvelopeDto`, `LogEventDto`, `LogAckDto` (JsonPropertyName-pinned lowercase wire form).
4. `ILogIngestionService` + `LogIngestionService` with `[ThreadStatic]` guard, dedup LRU, per-event try/catch, dropped→Warning emission.
5. `WebSocketLogProtocol.HandleFrameAsync` stateless helper (BYOWS surface).
6. `LogController` + `AddSunlightLogIngestion` extension.
7. `WebSocketLogHandler.RunAsync` + `MapSunlightLogIngestion` extension. Server-side feature complete.
8. `Test/Runtime/Sunlight.Logging.Server.Test` — 16 unit + WAF round-trip tests, all green.
9. `WebSocketLogSink` (callback-based, `maxRetry`, `HandleAck`/`NotifyDisconnected`, `RetryExhausted`/`DisconnectedWithPending` events) + 7 framework tests with `FakeWebSocketSendCapture` + `ManualWindowTimer` + injectable clock.
10. `WindowWebSocketTransport` optional helper (for consumers without an existing WS).
11. `FailoverLogSink` + 5 tests covering routing, mid-batch disconnect drain, RetryExhausted reroute, detach forwarding.
12. `LogSinkFactory.CreateFromBootstrap` reading `__nscriptLogEndpoint`/`__nscriptLogWsEndpoint` globals.
13. Adapter `Settings.LogEndpoint` + `Constants.LogEndpointStr` + `SettingsProvider` parsing.
14. `IngestHost` + `SerilogLoggerConfig` — Kestrel host that serves static assets + ingest endpoints on a single 127.0.0.1 port, Serilog `CompactJsonFormatter` writes to `LogEndpoint`.
15. Rewire `Runner`/`Executor`: when `LogEndpoint` set, boot `IngestHost`, navigate to `httpBase/index.html`; when null, keep today's path. **Includes the dogfood addendum**: Runner/Executor inject `ILogger`, emit MEL events for browser launch/test run start/per-test/run end on categories `SunlightTestAdapter.TestRunner` and `SunlightTestAdapter.Executor`.
16. `SunlightTestAdapter.Integration.Test` — E2E hermetic test that POSTs a browser envelope AND emits via the adapter logger, then asserts BOTH families of events land in the same JSONL file with the right `SourceContext`.
17. `docs/framework-logging.md` update: new `id` schema field, WS sink + BYOWS recipe, FailoverLogSink, LogSinkFactory, server-side surface (`ILogIngestionService` / `AddSunlightLogIngestion` / `MapSunlightLogIngestion` / `WebSocketLogProtocol`), adapter integration.
18. `_PackAdapterAssembly` bundles `Sunlight.Logging.Server.dll` + all Serilog DLLs under `build/net8.0/` — verified pack output contains all 6 DLLs.

## Test plan

- [x] `dotnet build NScript_Full.sln -c Debug` — succeeds, 0 errors.
- [x] `dotnet build NScript_Full.sln -c Release` — succeeds, 0 errors.
- [x] New unit tests: 16 server-side + 9 LogEventId + 7 WebSocketLogSink + 5 FailoverLogSink + 2 integration = **39 new tests**, all green.
- [x] No regressions in existing tests when run sequentially (`NScript.CLR.Test` has pre-existing parallelism flakiness unrelated to this PR — passes alone, exists on master too).
- [x] `dotnet pack` produces a nupkg with all ingest-host DLLs bundled under `build/net8.0/`.
- [x] Hermetic E2E test asserts adapter + browser MEL events both land in the same `tests.jsonl` file.

## Deviations from plan

None. All 18 v2 steps + the dogfood addendum landed as specified.


gautam-achieveai (commented):
**[Gautam's review bot]** Deep review of the Sunlight log-ingestion pipeline (#99).

## Summary
Well-structured 18-commit series implementing a full log-ingestion pipeline: client-side WebSocket/HTTP sinks with ACK + failover, server-side ASP.NET Core controller + WS handler, and integration with SunlightTestAdapter via in-process Kestrel. Architecture is clean with BYOWS escape hatch, testing is comprehensive, and back-compat is carefully preserved.

## Findings

### MEDIUM - Dedup not acking duplicates may cause indefinite retransmission
**LogIngestionService.cs:153-158** - When the dedup LRU detects a duplicate id, TryIngestOne returns false so the id is NOT included in the ack response. The comment says "the original ack already shipped" - but the client is *retransmitting* precisely because it did NOT receive the original ack. The current behavior means the client will keep retransmitting that event until either maxRetry is hit (dropping the event) or the LRU evicts the id (after which it'll be ingested as "new"). Consider returning true for duplicates so the ack reaches the client and stops the retransmit loop.

### MEDIUM - WebSocketLogHandler: OperationCanceledException not caught gracefully
**WebSocketLogHandler.cs:65** - When the cancellation token fires during ReceiveAsync, an OperationCanceledException propagates out of RunAsync into the endpoint handler. A catch (OperationCanceledException) when (ct.IsCancellationRequested) { } would make the shutdown path clean and avoid noisy unhandled-exception logs.

### LOW - Custom NullLogger duplicates framework type
**Runner.cs:282-295** - The adapter defines its own NullLogger class. Microsoft.Extensions.Logging.Abstractions (transitively referenced via Serilog.Extensions.Logging) already provides NullLogger.Instance with identical semantics.

### LOW - WebSocketLogSink.Flush exception path doesn't track failures
**WebSocketLogSink.cs:198** - When sendPayload throws, events have been moved to inFlight but the payload wasn't sent. The retry mechanism handles this correctly, but there's no visibility into transport failures.

## Verdict
**APPROVE** - Production-quality work with thoughtful design. The dedup-ack behavior is the most significant point worth discussing, but practical impact is mitigated by maxRetry (events won't retransmit forever - they'll be dropped after 3 retries and routed to HTTP via RetryExhausted).

Sources/Runtime/Sunlight.Logging.Server/Services/LogIngestionService.cs
```diff
@@ -0,0 +1,259 @@
...
+        /// per-event try/catch in <see cref="IngestAsync"/>.
+        /// </summary>
+        private bool TryIngestOne(LogEventDto evt)
+        {
+            if (!string.IsNullOrEmpty(evt.Id) && this.IsDuplicate(evt.Id!))
+            {
+                // Idempotent retry: do NOT ack the duplicate. Returning
+                // false means the id stays out of the ack list, which is
+                // what the client expects (the original ack already shipped).
+                return false;
```
gautam-achieveai: [Gautam's review bot] The dedup path returns false here, excluding the duplicate id from the ACK response. But the client retransmits because it never received the original ACK. Not acking duplicates means the client retransmits until maxRetry is exhausted. Consider returning true for duplicates - the event was already emitted to MEL, so acking is idempotent and stops unnecessary retransmission.

Sources/Runtime/Sunlight.Logging.Server/Hosting/WebSocketLogHandler.cs
```diff
@@ -0,0 +1,126 @@
...
+            var rented = ArrayPool<byte>.Shared.Rent(ReceiveBufferSize);
+            try
+            {
+                while (!ct.IsCancellationRequested && socket.State == WebSocketState.Open)
+                {
+                    // Reassemble one logical frame: keep reading slices
+                    // until EndOfMessage. We avoid allocating a MemoryStream
+                    // until we hit a fragmentation case, since the common
+                    // case is a single sub-buffer read.
+                    var first = await socket.ReceiveAsync(rented.AsMemory(0, ReceiveBufferSize), ct)
```
gautam-achieveai, line 65: [Gautam's review bot] When ct fires during ReceiveAsync, OperationCanceledException propagates out of RunAsync into the endpoint handler. Consider adding: catch (OperationCanceledException) when (ct.IsCancellationRequested) { } for a clean shutdown path without noisy exception logs in ASP.NET Core's request pipeline.

Sources/Runtime/Sunlight.Logging.Server/Services/LogIngestionService.cs
gautam-achieveai: [Gautam's bot] Fixed in 40da8600.

`TryIngestOne` now returns `true` on the dedup-hit branch (before any emit), so the duplicate id IS added to the ack list. The event was already emitted on first receipt, so re-emitting is still skipped — the skip-emit behaviour is unchanged. Acking the duplicate is what stops the client retransmit loop: the client only retransmits an id it never saw acked, so previously (`return false`) it would keep resending until `maxRetry` dropped the event. Acking again is idempotent on the client (it just removes the id from its in-flight map).

Updated coupled test `IngestAsync_DuplicateId_SkipsSecondEmit`: now expects `acked.Count == 2` with both entries == "dup-1", while `_provider.Logs.Count` stays 1 (emit still skipped). Server tests: 16/16 green.

Sources/Runtime/Sunlight.Logging.Server/Hosting/WebSocketLogHandler.cs
gautam-achieveai: [Gautam's bot] Fixed in 40da8600.

`RunAsync` now has a `catch (OperationCanceledException) when (ct.IsCancellationRequested)` between the receive loop and the `finally`. On a normal shutdown/abort the OCE is swallowed there so the endpoint pipeline no longer logs a noisy unhandled exception for the expected close path. The existing `finally { ArrayPool<byte>.Shared.Return(rented); }` still runs unconditionally, so the rented buffer is always returned. The filter (`when ct.IsCancellationRequested`) keeps any non-cancellation OCE propagating as a genuine error.

gautam-achieveai:
[Gautam's bot] Iteration 2 — also addressed the two LOW findings (no inline threads were open for these), pushed in 40da8600:

**LOW (Runner.cs duplicate NullLogger):** Removed the bespoke `internal sealed class NullLogger : ILogger` from `Sources/Compiler/SunlightTestAdapter/Runner.cs` and switched the default to `Microsoft.Extensions.Logging.Abstractions.NullLogger.Instance`. The MEL type resolves transitively (AspNetCore framework reference + Serilog.Extensions.Logging), so no new `PackageReference` was needed. Added `using Microsoft.Extensions.Logging.Abstractions;` to both `Runner.cs` and `Executor.cs`. Adapter builds clean.

**LOW (transport-failure visibility in WebSocketLogSink):** Added a diagnostic `sendFailureCount` field (init 0 in the ctor next to `droppedCount`), incremented inside both fire-and-forget catch blocks (the `Flush` send and the per-retry resend), exposed via `internal int SendFailureCount`. This is purely observational — the affected event still stays in-flight and is retransmitted, so delivery behaviour is unchanged. No console writes or new dependencies (this is NScript framework code that transpiles to JS). Added a focused unit test `TestSendFailureLeavesEventInFlightAndBumpsCounter` that throws from `sendPayload` and asserts the event stays in-flight and the counter increments.

Verification: server tests 16/16 green; framework suite 94/94 green (built with `-c Debug`); pre-commit hook ran the full solution build + compiler/E2E suites at 862/862 green. The two MEDIUM findings are addressed in their respective inline threads (now resolved).

gautam-achieveai (commented):
**[Gautam's review bot]** Incremental re-review (63421a6->40da860).

All 4 findings from the initial review have been addressed:

| # | Finding | Status |
|---|---------|--------|
| 1 | Dedup not acking duplicates (MEDIUM) | :white_check_mark: Fixed - TryIngestOne returns true on dedup hit; test updated |
| 2 | OperationCanceledException unhandled (MEDIUM) | :white_check_mark: Fixed - catch clause with when (ct.IsCancellationRequested) filter |
| 3 | Custom NullLogger duplicates framework type (LOW) | :white_check_mark: Fixed - removed, uses NullLogger.Instance from Abstractions |
| 4 | Flush exception path no visibility (LOW) | :white_check_mark: Fixed - sendFailureCount diagnostic counter + dedicated test |

Build clean, 16/16 server tests pass. Code quality remains high - good commit hygiene with each fix atomic and well-commented.

**Verdict: APPROVE** - no remaining concerns.

